PR #447, opened 2023-08-03
Fix/apro withdraw

nseguias:
We were simulating the exit pool using the total amount of shares the vault has, instead of only using UNBOND_QUEUE amount for slippage calculation


LaurensKubat (dismissed):
lgtm

LaurensKubat (approved):
lgtm

PR #467, opened 2023-08-16
Missing AddQueryFlagsToCmd for qVesting module on query "accounts"

magiodev:


arhamchordia (approved):
Looks good to me. Checked other modules as well. All of them have this flag.

PR #471, opened 2023-08-21
Adding PR template and Issues template.

arhamchordia:
Adding PR and issues opening templates.


magiodev (approved):
Looks good, we will refine those over time when we have some usage feedback

PR #472, opened 2023-08-21
Fix retry join pool

nseguias:
## 1. Overview
When manually executing a Retry for join pool, we were incorrectly transferring the total amount of BOND_QUEUE & FAILED_JOIN_QUEUE. This doesn't make sense as failed bonds were already transferred and sitting idle in the contract. This is now being handled correctly and we're only transferring BOND_QUEUE total amount.

## 2. Implementation details
fold_bonds was adding BOND_QUEUE and FAILED_JOIN_QUEUE, and returned the total of these two both. I removed FAILED_JOIN_QUEUE from the total calculation as we do not want to transfer those funds. Also fixed current tests to accommodate changes.

## 3. How to test/use
cd smart-contracts && cargo test 

## 4. Checklist

## 5. Limitations (optional)
When we call Retry, we also want to call Bond with some tokens to guarantee sending a join_pool transaction as this is only called in the handle_transfer_ack function.

### DIFF
--- a/smart-contracts/contracts/lp-strategy/src/execute.rs
+++ b/smart-contracts/contracts/lp-strategy/src/execute.rs
@@ -805,18 +805,8 @@ mod tests {
         let res = handle_icq_ack(deps.as_mut().storage, env, to_binary(&ibc_ack).unwrap()).unwrap();
 
         // we do NOT transfer any token here, failed bonds were already transferred to the contract before failing and stay there
-        match &res.messages[0].msg {
-            CosmosMsg::Ibc(IbcMsg::Transfer { amount, .. }) => {
-                assert_eq!(
-                    amount,
-                    &Coin {
-                        denom: "ibc/local_osmo".to_string(),
-                        amount: Uint128::zero(),
-                    }
-                );
-            }
-            _ => panic!("unexpected message type"),
-        }
+        // as we do not have any bond_queue items, we return None here
+        assert!(res.messages.is_empty());
 
         // check that the failed join queue is emptied
         assert!(FAILED_JOIN_QUEUE.is_empty(&deps.storage).unwrap());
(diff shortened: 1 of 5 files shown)


smart-contracts/contracts/lp-strategy/src/execute.rs
```diff
@@ -1027,28 +1008,19 @@ mod tests {
...
-            acc + amount
-        });
-
         // get the pending bonds total amount
         let pending_total_amount = pedning_bonds
             .iter()
             .fold(Uint128::zero(), |acc, bond| acc + bond.amount);
 
-        // check that the res amount matches the amount in both queues
+        // check that the res amount matches the amount in the pending queue ONLY
```
LaurensKubat, line 1016: We should also verify that the failed bond queue is properly emptied

smart-contracts/contracts/lp-strategy/src/proptests.rs
```diff
@@ -271,7 +261,7 @@ mod tests {
                             amount,
                             &Coin {
                                 denom: "ibc/local_osmo".to_string(),
-                                amount: failed_total_amount + pending_total_amount,
```
LaurensKubat, line 274: We should also verify here that the failed bonds are treated properly

LaurensKubat (changes requested):
We now support 0 sends with this, but we are still incorrectly summing the total to send if we do a retry, we are summing [here](https://github.com/quasar-finance/quasar/blob/605809b87f44686469a0628efa05861dc0735709/smart-contracts/contracts/lp-strategy/src/bond.rs#L140C9-L140C14)
How to test should also include some note on E2E tests. I believe we might have some in @magiodev s branch already

smart-contracts/contracts/lp-strategy/src/execute.rs
nseguias: we already do this a few lines below:
```
        // check that the failed join & pending queues are emptied
        assert!(FAILED_JOIN_QUEUE.is_empty(&deps.storage).unwrap());
        assert!(PENDING_BOND_QUEUE.is_empty(&deps.storage).unwrap());
```

smart-contracts/contracts/lp-strategy/src/proptests.rs
nseguias: done

LaurensKubat:
merged into https://github.com/quasar-finance/quasar/pull/429

PR #473, opened 2023-08-21
CI Fixes

arhamchordia:
## 1. Overview

This PR aims to fix the CI issues and updations of versions.

## 2. Implementation details

- updates go version to 1.20.
- changes the build CI that keeps on running on every commit even if the related files are not changed.
- earlier the CI on the commits kept on running even if we added new commit which was adding to the increase costs. Now the new commits cancels out all the CI running on the previous commit on the same branch.
- CI for go runs on changes in go related files.
- CI for rust runs on changes in rust related files.

## 3. How to test/use

- Testing happens on this PR's CI.

## 4. Checklist

No readme in this case.

## 5. Limitations (optional)

- Limitation is that the Darwin build is not included. It take a lot of time to build it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,6 +31,7 @@ jobs:
         uses: actions/checkout@v3
       - name: Get git diff
         uses: technote-space/get-diff-action@v6.1.2
+        if: env.GIT_DIFF
         with:
           PATTERNS: |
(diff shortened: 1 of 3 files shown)


akure (approved):

magiodev (approved):
Looks good to me

PR #488, opened 2023-09-07
Feat/cl vault range tests

LaurensKubat:
## 1. Overview

- We now allow deposits with only 1 assets via exact deposit. If the position needs more assets, the deposit will be rejected there
- added a swap to the test tube tests so that the current tick is moved inside the range
- added initial tests for range modifications
- fixed a bug in the merge steps
- fixed some halborn bugs
- Changed modifyRange api

## 4. Checklist


LaurensKubat:
Merging since we are merging to the feat/cl-vault branch

PR #491, opened 2023-09-11
Fix/cl vault swap calc

LaurensKubat:
## 1. Overview

changes the swap math according to https://atiselsts.github.io/pdfs/uniswap-v3-liquidity-math.pdf

## 2. Implementation details

Our swap math substitutes P_l and P_u with P_c in the wrong places. This fixes it. See the above link for the math logic

## 4. Checklist


njerschow (approved):
lgtm

PR #495, opened 2023-09-11
add some fixes and checks for rewards

LaurensKubat:
## 1. Overview

- Fixes and a check in instantiate
- changes the percentage and sub_percentage functions to ratio and sub_ratio which work on decimals directly
- readds some tests and some initial work on the test helpers

## 4. Checklist


njerschow (approved):
lgtm

PR #496, opened 2023-09-12
Feat/cl vault swap test

njerschow:
## 4. Checklist


smart-contracts/contracts/cl-vault/src/vault/range.rs
magiodev, line 1: I think with this helpers.rs math we should go back to > and > only, not equal
```
if (amount0.is_zero() && modify_range_state.lower_tick <= pool_details.current_tick)
        || (amount1.is_zero() && modify_range_state.upper_tick >= pool_details.current_tick)
    {
```

njerschow: @LaurensKubat any insight here?
@magiodev why?

PR #503, opened 2023-09-16
use modify_range msg slippage instead of vault_config

LaurensKubat:
## 1. Overview

Changes from where slippage is passed when swapping for the second position

## 4. Checklist


njerschow (approved):
nice, lgtm - the only thing I would additionally suggest would be to remove the swap_max_slippage from the vault_config. But that requires additional migrate logic. We can leave it for cleanup at a later point

LaurensKubat:
Added a clickup task for the config item

PR #506, opened 2023-09-19
Feat/cl vault prop assertions

magiodev:
## 1. Overview

Implementing assertions to CL Vault contract proptests.

## 2. Implementation details

Asserting before and after values retrieved by chain modules / contract using assert macros from Rust testing crates.

## 3. How to test/use

Run the test inside contracts/cl-vault/test_tube/proptest.rs

## 4. Checklist

## 6. Future Work (optional)

Creating more deterministic (less randomic) test cases like forcing the current tick to be in-range, and so on.

### DIFF
--- a/smart-contracts/contracts/cl-vault/src/test_tube/proptest.rs
+++ b/smart-contracts/contracts/cl-vault/src/test_tube/proptest.rs
@@ -1,5 +1,7 @@
 #[cfg(test)]
 mod tests {
+    use std::str::FromStr;
+
     use cosmwasm_std::{assert_approx_eq, Addr, Attribute, Coin, Decimal, Uint128};
     use osmosis_std::types::cosmos::bank::v1beta1::{QueryBalanceRequest, QueryBalanceResponse};
     use osmosis_std::types::cosmwasm::wasm::v1::MsgExecuteContractResponse;
@@ -14,7 +16,6 @@ mod tests {
     };
     use proptest::prelude::*;
 
-    use crate::debug;
     use crate::{
         helpers::sort_tokens,
         math::tick::tick_to_price,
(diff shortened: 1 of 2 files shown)


smart-contracts/contracts/cl-vault/src/test_tube/proptest.rs
```diff
@@ -408,43 +457,20 @@ mod tests {
...
     }
-    */
-
-    // COMPOSE STRATEGY
 
     // get_initial_range generates random lower and upper ticks for the initial position
     prop_compose! {
-        // TODO: evaluate if lower_tick and upper_tick are too much arbitrary
-        fn get_initial_range()(lower_tick in 1i64..1_000_000, upper_tick in 1_000_001i64..2_000_000) -> (i64, i64) {
+        fn get_initial_range()(lower_tick in -1_000_000i64..1_000_000, upper_tick in 1_000_001i64..2_000_000) -> (i64, i64) {
```
LaurensKubat, line 473: These are still abritrary ticks right? I think we want to use MIN_TICK and MAX_TICK from math/tick.rs here

smart-contracts/contracts/cl-vault/src/test_tube/proptest.rs
```diff
@@ -408,43 +457,20 @@ mod tests {
...
+    // COMPOSE STRATEGY
 
-    fn assert_update_range() {
-        todo!()
+    // Actions enum
+    #[derive(Clone, Copy, Debug)]
+    enum Action {
+        Deposit,
+        Withdraw,
+        // Swap,
```
LaurensKubat, line 467: Why are we excluding swaps right now?

LaurensKubat (commented):
Some question on the current tests, but the assertions look good

PR #507, opened 2023-09-21
change min spot price to 10^-18

LaurensKubat:
## 1. Overview

Changes the min spot price according to the [osmosis docs](https://docs.osmosis.zone/osmosis-core/modules/concentrated-liquidity/#geometric-tick-spacing-with-additive-ranges)

## 2. Implementation details

Change of a single constant
 
## 3. How to test/use

All current tests should provide sufficient coverage

## 4. Checklist

- [ x] Does the Readme need to be updated?
Current README does not cover tick math


nseguias (commented):
There might be a problem with this approach if we want to support quote assets with 6 decimals and base assets with 18 in the future, at least for assets with very low prices.

You can read about it if you search for `Precision Issues With Price` in the osmosis doc you referred to in the overview.

I think I got the min spot price equal to 10^-12 from there.

Also, PR overview says `max` instead of `min`.

PR #508, opened 2023-09-21
Fix/tick to price cache usage

LaurensKubat:
## 1. Overview

Build the tick exp cache in instantiate instead of in `tick_to_price`, and make the tests runs in ci

## 2. Implementation details

Call `build_tick_exp_cache` in instantiate instead of `tick_to_price`

## 3. How to test/use

Current tests should cover this. The behaviour of tick_to_price should not change

## 4. Checklist

- [ x] Does the Readme need to be updated?
No


nseguias (dismissed):
LGTM, after we fix the failing tests & lint, adding `build_tick_exp_cache(storage).unwrap()` on each failing test should be enough

magiodev (approved):
LGTM. Also tests are passing on my local and CI looks fine too

PR #511, opened 2023-09-28
Fix/balance queries

njerschow:
## 1. Overview

Balance queries

## 2. Implementation details

Fix balance queries 

## 3. How to test/use

check balance

## 4. Checklist

- [ ] Does the Readme need to be updated?no


njerschow (commented):
Should we be excluding all other tokens from unused_balance? Currently the third party reward tokens and the vault token also come back

LaurensKubat (dismissed):
lgtm

LaurensKubat (approved):
lgtm

PR #515, opened 2023-10-08
add partial swap test to cl tests

LaurensKubat:
## 1. Overview

Adds a test for partial swaps, this test currently does not come with assertions, will add this later, but we want to check this test manually now

## 4. Checklist


nseguias (approved):
Test lgtm. How does block time or block height work in here to calculate TWAP?

LaurensKubat:
> Test lgtm. How does block time or block height work in here to calculate TWAP?

Since there are no price changes, in those blocks, the different block height should not affect our TWAP, see osmosis docs for details

njerschow (approved):
lgtm

PR #540, opened 2023-11-13
`error`: overflow in sum

ajansari95:


njerschow (approved):
I don't see an issue with these changes however I am confused why we would ever hit an overflow on testnet during migration, seeing as we always set a valid Uint128 value for the vault SHARES. so, how would there ever be overflow unless our minting code is extremely wrong.

PR #550, opened 2023-11-23
Range Automation Main

njerschow:
## 4. Checklist


smart-contracts/contracts/cl-vault/src/query.rs
```diff
@@ -112,6 +113,15 @@ pub fn query_metadata(deps: Deps) -> ContractResult<MetadataResponse> {
     })
 }
 
+pub fn query_dex_router(deps: Deps) -> ContractResult<Option<String>> {
+    let dex_router = DEX_ROUTER.may_load(deps.storage)?;
```
magiodev, line 117: Could this be a .load()?; as the underlying state item is not an Option? Below code would be cleaner too.

smart-contracts/contracts/range-middleware/NOTICE
```diff
@@ -0,0 +1,13 @@
+Copyright 2023 Nikita (rizz) <nj18@nyu.edu>
```
magiodev, line 1: I dont know if we should update this

magiodev (commented):
Just some minor comment, lgtm.

smart-contracts/contracts/range-middleware/NOTICE
njerschow: lol probably

smart-contracts/contracts/cl-vault/src/query.rs
njerschow: if we use load without dex router set then this query will error

magiodev: but in what case we would able to do so? the migrate() entrypoint sets this item at state. querying before migration is unfeasible. or there is a scenario where we migrate the code without executing the migrate() entrypoint that im missing?

smart-contracts/contracts/cl-vault/src/vault/range.rs
```diff
@@ -417,15 +423,42 @@ pub fn handle_swap_reply(
 ) -> Result<Response, ContractError> {
     // TODO: Remove handling of data. if we keep reply_on_success in the caller function
     match data.clone() {
-        SubMsgResult::Ok(_msg) => handle_swap_success(deps, env, data.try_into()?),
+        SubMsgResult::Ok(msg) => {
+            let resp: Result<MsgSwapExactAmountInResponse, _> = data.try_into();
+            let tokens_out: Result<String, ContractError> = match resp {
+                Ok(msg) => Ok(msg.token_out_amount),
+                Err(_) => {
```
LaurensKubat, line 430: I don't fully understand why we are doing the advanced handling here, I believe Cosmwasm doesn't return events from CosmosSdk messages anyway in versions < v2.0. Are we expecting cases where Osmosis returns we are unable to handle the swap response and we don't want to revert the tx in that case?

smart-contracts/contracts/range-middleware/.circleci/config.yml
LaurensKubat, line 1: We can remove this file since it's template CI from the Cosmwasm template and we're not using CircleCI

LaurensKubat (approved):
in general lgtm, some minor things I don't fully understand

smart-contracts/contracts/cl-vault/src/helpers.rs
```diff
@@ -452,6 +452,23 @@ pub fn get_liquidity_amount_for_unused_funds(
     Ok(max_initial_deposit_liquidity.checked_add(post_swap_liquidity)?)
 }
 
+pub fn extract_attribute_value_by_ty_and_key(
```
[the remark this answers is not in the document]
LaurensKubat: After looking through the entire PR, I saw this used at the swap result handling, if we might want to keep it depending on the discussion of that handling

smart-contracts/contracts/cl-vault/src/vault/range.rs
LaurensKubat: If we're trying to handle different swap types here, we probably want to use an untagged struct instead and deserialize against that

smart-contracts/contracts/cl-vault/src/helpers.rs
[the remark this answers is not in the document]
magiodev: adjusted as suggested

magiodev (approved):
lgtm after all changes

PR #556, opened 2024-01-03
migrate old invest into new invest for OPRO specifically

njerschow:
## 1. Overview

Somehow our invest field in opro never landed in new_invest which results in it returning null. Hard to say why this only happened after the most recent migration and not before on OPRO and not APRO, but I suspect we hadnt migrated these vaults together.

see raw query for `invest` is there with: 696E76657374
and for `new_invest` it is null with: 6E65775F696E76657374


arhamchordia (approved):
Looks good to me.
I feel it was mostly because of dependency changes as this is the only thing we changed in the Basic vault code.

PR #562, opened 2024-01-16
Feat/multiple position support

LaurensKubat:
Todo I have to write a good Pr description here
## 1. Overview

This PR adds multiple position support or multi range to our vaults.
This implementation supports:
- Multiple ranges with different amounts of funds in them
- direct deposits of user assets into certain ranges
- Separation of in which positions the user deposits from the amount of funds in the positions

## 2. Implementation details

The changes in this PR can be split up into 3 parts:
1. The multiple range operation part
2. user deposits
3. user withdraws

### 1. Multiple range operation
to operate the multiple ranges, this vault exposes multiple endpoints, namely:
`MovePosition`, this moves a position between ranges and swaps any free funds to fit that range, similar to the previous singe position range `ModifyRange` interface
`AddRatio`. This increase the ratio of a position
`LowerRatio` This decreases the ratio of a position
`IncreaseFunds` This increases the funds in a position without swapping
`DecreaseFunds` This withdraws funds from a position
`CreatePosition` This creates a new position
`DeletePosition` This deletes and old position

Withdrawing funds from a positon is explicitly made a different call than delete position. the range admin is not allowed to withdraw all liquidity via `DecreaseFunds`, but should be done through `DeletePosition`

### 2. User Deposits
This PR deposits user funds into each position according to the ratio, in the end to account for an exact user deposit into all positions, we need to calculate the combination of how many funds should go to a single position out of the set of multiple positions, and the positions internal ratio between the two assets it accepts.
For example, we have the following positions
```
Position 1:
10 osmo
90 atom
ratio: 10

Position 2:
50 osmo
50 atom
ratio: 5
```
So 2/3 of our funds should go into position 1 and 1/3 into position two, naively depositing 2/3 of our osmo and atom in position one and the rest into position two would be very inefficient. we instead want to figure out what % of all incoming osmo should go to position 1 and what % should got to position 2.
given that 2/3 need to go to position 1, we can allocate 66osmo, we can also allocate 66 atom to position 1. We already know that not all osmo will be used here, instead we calculate the effective ratio of position 1 as 1/10*2/3 osmo and 9/10*2/3 atom, so 2/30 and 18/30. Meaning that 2/30 and 18/30 of all assets should go to this position
Doing the same for position 2 gives us a ration there 1/2*1/3 and 1/2 * 1/3 , or 1/6 and 1/6 or 5/30 and 5/30 (note we keep the denominator here since we are using fraction instead of percentages, we do not expect a total of 1)
now we know that we want 7 osmo for each 23 atom and to allocate 2 of that osmo to position 1 and 5 to position 2

### 3. User Withdraws
This PR contains a very simple and basic implementation of withdraw logic. A user owns a ratio of shares compared to the vaults total shares, meaning we can just withdraw the users percentage from each position and the free funds. This can potentially be improved in the future to always have some sort of setting/ prefererred position to get the funds from

## 3. How to test/use

Most tests exist in the tests folder and are written through test-tube

## 4. Checklist

## 5. Limitations (optional)
- Any action that requires split up swaps cannot easily be chained in the same transaction with any other actions. Unknown at this time whether that is a needed action and how the specs should work to see if we should move that to t

## 6. Future Work (optional)

- The withdraw implementation might cause more work for the strategists combined with Osmosis uptime based incentives. An improvement here would be to add some sort of preferential withdraw location system, such as first try to withdraw the user value from the free balance, then try to withdraw user value from a predetermined position and else go for all positions
- Zapping, could be added in this case by taking all funds a user wants to deposit and letting any unusable funds lie in unused balance


smart-contracts/contracts/cl-vault/Cargo.toml
```diff
@@ -49,7 +49,7 @@ cw-dex = "0.1.1"
 cw-dex-router = {version = "0.1.0", features = ["library"]}
 cw-storage-plus = "1.0.1"
 cw-utils = "0.16.0"
-cw-vault-multi-standard = {git = "https://github.com/quasar-finance/cw-vault-standard", features = ["lockup", "force-unlock"]}
+cw-vault-multi-standard = {git = "https://github.com/quasar-finance/cw-vault-standard", features = ["lockup", "force-unlock"], branch = "feat/change-convert-assets-return-type"}
```
akure, line 52: Cargo build fails with error,
error: failed to select a version for the requirement `osmosis-test-tube = "^17.0.0-rc0"`
candidate versions found which didn't match: 22.0.0
I think this version 17.0.0-rc0 does not exist any more

smart-contracts/contracts/cl-vault/src/math/liquidity.rs
```diff
@@ -41,16 +41,71 @@ pub fn _liquidity0(
     let intermediate_2 = Decimal256::from_atomics(intermediate, 36).unwrap();
 
     // since we start with Decimal and multiply with big_factor, we expect to be able to convert back here
-    Ok(Decimal::new(intermediate_2.atomics().try_into()?))
+    Ok(Decimal256::new(intermediate_2.atomics()))
+}
+
+/// asset0 calculates the
```
akure, line 47: There is a lot of common code in the methods of asset0, asset1, and _liquidity0/1 that can be refactored to reuse common parts by using simpler helpers.

smart-contracts/contracts/cl-vault/src/math/liquidity.rs
```diff
@@ -41,16 +41,71 @@ pub fn _liquidity0(
...
+pub fn asset0(
```
akure, line 48: unit test cases are missing.

smart-contracts/contracts/cl-vault/src/helpers.rs
```diff
@@ -231,13 +263,17 @@ pub fn assert_admin(deps: Deps, caller: &Addr) -> Result<Addr, ContractError> {
...
+        return 0; // Return 0 when amount is less than multiple
+    }
+
     let remainder = amount % multiple;
     if remainder == 0 {
-        amount
+        amount // Amount is already a multiple of 'multiple'
     } else if amount < 0 {
-        amount - remainder
+        amount + remainder.abs() // Round down to the nearest multiple
```
akure, line 274: This also looks effectively round up if amount < 0, as we are adding remainder.abs()  to a negative number.

smart-contracts/contracts/cl-vault/src/helpers.rs
```diff
@@ -290,139 +326,185 @@ pub fn get_max_utilization_for_ratio(
...
-    if p.position.is_none() {
-        return Ok(Decimal256::zero());
+// TODO, allocating funds gives a ratio for each position, we should theb
+pub fn allocate_funds_per_position(
+    _deps: DepsMut,
+    positions: Vec<(Position, FullPositionBreakdown)>,
+    spot_price: Decimal,
+    token0: Uint128,
+    token1: Uint128,
+) -> Result<Vec<(Position, Uint128, Uint128)>, ContractError> {
```
akure, line 336: Edge case can be handled in the beginning in the func, 
1. when token0 or token1 is zero
2. When positions is empty. 
3. When spot price is zero to represent bad/rugged  pool situation.

smart-contracts/contracts/cl-vault/src/helpers.rs
```diff
@@ -290,139 +326,185 @@ pub fn get_max_utilization_for_ratio(
...
+/// within the vault
+fn get_min_ratio_per_position(
+    positions: Vec<(Position, FullPositionBreakdown)>,
+    spot_price: Decimal,
+) -> Result<Vec<(Position, PositionRatio)>, ContractError> {
+    // per position, calculate the ratio of asset1 and asset2 a position needs
+
+    // for each position, we take some set amount of liquidity and calculate how many asset0 and asset1 those would give us
+    // we assume
+    let liquidity: Decimal256 = Decimal::new(100_000_000_000_000_000_000_000_u128.into()).into();
```
akure, line 386: Couldn't get the logic for using 100_000_000_000_000_000_000_000_u128 number here for liquidity.

smart-contracts/contracts/cl-vault/src/helpers.rs
akure, line 336: Haven't checked yet, but does this normalized approach gives more allocation to narrow range position compared to wider range position?

smart-contracts/contracts/cl-vault/src/vault/deposit.rs
```diff
@@ -36,90 +41,218 @@ pub(crate) fn _execute_any_deposit(
...
 
-    // Create coins_to_send with no zero amounts
-    let mut coins_to_send = vec![];
-    if !token0.amount.is_zero() {
-        coins_to_send.push(token0.clone());
-    }
-    if !token1.amount.is_zero() {
-        coins_to_send.push(token1.clone());
-    }
+    let position_funds = psf.iter().zip(positions);
```
akure, line 73: Could this be optmized in some way, as type of position_funds would become,  Zip<Iter<(Position, Uint128, Uint128)>, IntoIter<(Position, FullPositionBreakdown)>, where Position comes two times and position id three time in the zip.
Also we can already make sure that zipped values of two tuple  (Position, Uint128, Uint128), (Position, FullPositionBreakdown)  are already sorted and have same position id on same index.

smart-contracts/contracts/cl-vault/src/helpers.rs
```diff
@@ -290,139 +326,185 @@ pub fn get_max_utilization_for_ratio(
...
+    spot_price: Decimal,
+) -> Result<Uint128, ContractError> {
+    let total = token0
+        .checked_add(token1.multiply_ratio(spot_price.denominator(), spot_price.numerator()))?;
+
+    Ok(total)
+}
+
+#[cw_serde]
+pub struct PositionRatio {
```
akure, line 437: I think asset0 and asset1 should be Uint256. In their multiply impl, it might easily overflow.

akure: Also you have used full_mul which returns Uint256 in checked_mul impl.

smart-contracts/contracts/cl-vault/src/helpers.rs
```diff
@@ -290,139 +326,185 @@ pub fn get_max_utilization_for_ratio(
...
+    spot_price: Decimal,
+) -> Result<Vec<(Position, PositionRatio)>, ContractError> {
+    // per position, calculate the ratio of asset1 and asset2 a position needs
+
+    // for each position, we take some set amount of liquidity and calculate how many asset0 and asset1 those would give us
+    // we assume
+    let liquidity: Decimal256 = Decimal::new(100_000_000_000_000_000_000_000_u128.into()).into();
+    let ps: Result<Vec<(Position, PositionRatio)>, ContractError> = positions
+        .into_iter()
+        .map(|(p, fp)| {
```
akure, line 389: The Variable name "Position" makes the readability difficult as it is also a local to vault state.rs and  as well as osmosis std concentrated liquidity beta1.rs, so here p and pos both are Positions variable. Can we think of improving local names here?

smart-contracts/contracts/cl-vault/src/helpers.rs
```diff
@@ -290,139 +326,185 @@ pub fn get_max_utilization_for_ratio(
...
+
+    // now that we know how much tokens a positions internal ratio needs, we need to normalize these internal ratios to eachother using the positions ratios
+    // Each position might get ratio/total_ratio of tokens. We want to find the effective ratio for each position then.
+    // we multiply the position's internal ratio by the positions external ratio
+    // The external ratio is the positions ratio divided by the total ratio of all positions
+    let positions: Result<Vec<(Position, PositionRatio)>, ContractError> = positions
+        .into_iter()
+        .map(|(p, internal_ratio)| {
+            let external_ratio = Decimal::from_ratio(p.ratio, total_ratio);
+            let total_ratio = internal_ratio.checked_mul_ratio(external_ratio)?;
```
akure, line 417: Should we call this effective ratio after multiplying internal and external to distinguish it with previously calculated total_ratio.

smart-contracts/contracts/cl-vault/src/vault/range/create_position.rs
```diff
@@ -0,0 +1,69 @@
...
+        .add_attribute("upper_price", upper_price.to_string()))
+}
+
+pub fn handle_range_new_create_position(
+    deps: DepsMut,
+    result: SubMsgResult,
+) -> Result<Response, ContractError> {
+    let v: MsgCreatePositionResponse = result.try_into()?;
+
+    let ratio = CURRENT_RATIO.load(deps.storage)?;
```
akure, line 58: shall we clean up CURRENT_RATIO , after saving the ratio with position id in the POSITIONS store.

smart-contracts/contracts/cl-vault/src/vault/range/modify_percentage.rs
```diff
@@ -0,0 +1,164 @@
...
+    helpers::{get_one_or_two, get_unused_balances},
+    msg::{ExecuteMsg, MergePositionMsg},
+    reply::Replies,
+    rewards::CoinList,
+    state::{Position, CURRENT_POSITION_ID, POOL_CONFIG, POSITIONS},
+    vault::concentrated_liquidity::{create_position, get_position, withdraw_from_position},
+    ContractError,
+};
+
+/// increase the ratio that a position has in the vault for depositors
```
akure, line 16: What is the flow for add ratio and lower ratio, these two entry points are only updating the ratio value in the POSITION store. When it is supposed to reflect in the pool?

smart-contracts/contracts/cl-vault/src/vault/range/move_position.rs
```diff
@@ -321,11 +316,16 @@ pub fn do_swap_deposit_merge(
     } else {
         // if we have not tokens to swap, that means all tokens we correctly used in the create position
         // this means we can save the position id of the first create_position
-        POSITION.save(
+        // if position_id is None, then no first position was ever created, meaning we should never hit this else case.
+        let position_id = position_id.unwrap();
+        let modify = MODIFY_RANGE_STATE.load(deps.storage)?.unwrap();
```
akure, line 321: Idea - 
As we are now dealing with multi-range, Can we change MODIFY_RANGE_STATE internal struct ModifyRangeState to have current position id field and MODIFY_RANGE_STATE as vector of ModifyRangeState. That will help if we want to modify range for multiple positions at once.

smart-contracts/contracts/cl-vault/src/vault/range/move_position.rs
```diff
@@ -473,16 +473,24 @@ pub fn handle_iteration_create_position_reply(
         .target_range_position_ids
         .push(create_position_message.position_id);
 
+    let mrs = MODIFY_RANGE_STATE.load(deps.storage)?.unwrap();
```
akure, line 476: Should we also clean the MODIFY_RANGE_STATE, after its use is finished after merge.

smart-contracts/contracts/cl-vault/src/contract.rs
```diff
@@ -140,6 +163,8 @@ pub fn reply(deps: DepsMut, env: Env, msg: Reply) -> Result<Response, ContractEr
         Replies::RangeIterationCreatePosition => {
             handle_iteration_create_position_reply(deps, env, msg.result)
         }
+        Replies::RangeAddToPosition => handle_add_to_position_reply(deps, env, msg.result),
```
akure, line 166: for naming consistency : 
handle_add_to_position_reply  -> handle_range_add_to_position_reply

smart-contracts/contracts/cl-vault/src/helpers.rs
```diff
@@ -290,139 +326,185 @@ pub fn get_max_utilization_for_ratio(
...
+
+    let total_ratio = positions
+        .iter()
+        .fold(Uint128::zero(), |acc, (p, _)| acc + p.ratio);
+
+    // now that we know how much tokens a positions internal ratio needs, we need to normalize these internal ratios to eachother using the positions ratios
+    // Each position might get ratio/total_ratio of tokens. We want to find the effective ratio for each position then.
+    // we multiply the position's internal ratio by the positions external ratio
+    // The external ratio is the positions ratio divided by the total ratio of all positions
+    let positions: Result<Vec<(Position, PositionRatio)>, ContractError> = positions
```
akure, line 413: Can we write a small helper for doing this ratio related math, and use the helper like get_final_ratio(...), or get_effective_ratio(...) . The idea is to do simulate and see how ratios works from internal_ratio  to total_ratio on multiplying with the external ratio in unit tests.

smart-contracts/contracts/cl-vault/src/vault/deposit.rs
```diff
@@ -36,90 +41,218 @@ pub(crate) fn _execute_any_deposit(
...
+    let msgs: Result<Result<Vec<SubMsg>, ContractError>, ContractError> = position_funds
+        .map(|((pos1, amount0, amount1), (pos2, fp))| {
+            if pos1.position_id != pos2.position_id {
+                // this would be a programming error
+                panic!("Position ids in the zipped iterator don't match")
+            }
+
+            // we want to make sure deposit 1 is item 1 in the queue, deposit 2 item 2 in the queue
+            // TODO can this be less complex?
+            CURRENT_DEPOSIT.push_back(
```
akure, line 88: Current use of CURRENT_DEPOSIT look works, but for clarity we can also use one more state variable for pushing back the updated deposit in the handle_deposit_create_position_reply, which is finally used in the mint operation.

smart-contracts/contracts/cl-vault/src/vault/deposit.rs
```diff
@@ -36,90 +41,218 @@ pub(crate) fn _execute_any_deposit(
...
+            token1.amount - total_allocated.1,
+        ),
     )?;
 
     Ok(Response::new()
-        .add_submessage(SubMsg::reply_on_success(
-            create_position_msg,
-            Replies::DepositCreatePosition as u64,
-        ))
+        .add_submessages(msgs??)
```
akure, line 142: Question - what is the mechanism when one of the sub msg failed here?

smart-contracts/contracts/cl-vault/src/math/liquidity.rs
LaurensKubat: Agreed, but this way it's a 1:1 copy of the CL formulas from the [Osmosis docs](https://github.com/osmosis-labs/osmosis/tree/main/x/concentrated-liquidity), refactoring would require a deeper understanding of the actual formulas when checking whether we implement the osmosis math

smart-contracts/contracts/cl-vault/src/helpers.rs
LaurensKubat: I'm using some big number so we don't have rounding errors, we want to take the same amount for each position to get some amount out, if we want we can lower it a bit anyway, but we want to be in atleast the 1000's to 100_000 tokens range imo

smart-contracts/contracts/cl-vault/src/helpers.rs
LaurensKubat: Good question, will try to find out

smart-contracts/contracts/cl-vault/src/vault/deposit.rs
LaurensKubat: > Also we can already make sure that zipped values of two tuple (Position, Uint128, Uint128), (Position, FullPositionBreakdown) are already sorted and have same position id on same index.
This is already checked in
```
            if pos1.position_id != pos2.position_id {
                // this would be a programming error
                panic!("Position ids in the zipped iterator don't match")
            }
```
We never hit this panic with the current code
Sorting again is not needed right now so would be just extra work

LaurensKubat: >Could this be optmized in some way, as type of position_funds would become, Zip<Iter<(Position, Uint128, Uint128)>, IntoIter<(Position, FullPositionBreakdown)>, where Position comes two times and position id three time in the zip.

We could minimize this in a different struct, It would not lower the copy load afaik, but would be better legible. What do you think would work best here?

smart-contracts/contracts/cl-vault/src/helpers.rs
LaurensKubat: Heard, will try to improve somehow, the thing is they are all the same thing generally, but just with different scope. Any comments on how to improve here?

smart-contracts/contracts/cl-vault/src/helpers.rs
LaurensKubat: Good one

LaurensKubat: changed in [bc4925c](https://github.com/quasar-finance/quasar/commit/bc4925c356621b7777e9fafdaec620baab63f86a)

smart-contracts/contracts/cl-vault/src/vault/range/modify_percentage.rs
LaurensKubat: Yes, it is only supposed to be reflected in user the ratio users deposit it, we want to give the strategists the freedom to separate the wanted incoming funds of the vault vs the actual funds in the pool

smart-contracts/contracts/cl-vault/src/vault/range/move_position.rs
LaurensKubat: I think it's less complex to try to keep all actions only running on one position at a time, if we need the capability to move multiple at once, we can run the multiple execute msges in the same transaction

smart-contracts/contracts/cl-vault/src/vault/range/move_position.rs
LaurensKubat: Yes, will add this

LaurensKubat: changed in 8b2ffc162231d16d6d43e531728ab02ca4f69ced

smart-contracts/contracts/cl-vault/src/contract.rs
LaurensKubat: renamed in 6506fbf7a0b6aa806ae10f36c511b2961490ab25

smart-contracts/contracts/cl-vault/src/vault/deposit.rs
LaurensKubat: All the submsges are reply on success, so if a submsg fails, we don't get a callback but get regular error behaviour instead, meaning the entire transaction is reverted

smart-contracts/contracts/cl-vault/src/helpers.rs
LaurensKubat: Yes, will write some helpers, good call

LaurensKubat: done in 262cdfb0f156e596c351bff22d15d8a078f5623e

smart-contracts/contracts/cl-vault/README.md
```diff
@@ -155,3 +155,41 @@ $$K=\sqrt{P_c}\sqrt{P_l}\frac{\sqrt{P_u}-\sqrt{P_{c}}}{\sqrt{P_{c}}-\sqrt{P_l}}$
...
 We can actually create further optimizations here by increasing the amount of tokens to swap by half of the expected slippage, which will lead to returning (not being able to deposit) less tokens.
+
+# Multi-position support
+## intro
+The CL vault supports multiple positions internally. 
+
+### position ratios
+Each position is saved in the contract with a ratio. This ratio dictates how many tokens are sent to a position when a user deposits.
+
+Given any current spot price, each position needs a different amount of tokens. Lets take 3 positions $P_1$, $P_2$ and $P_3$. at some price $P$, $P_1$ needs 30% of token0 and 70% of token1. $P_2$ needs 45% token0 and 55% token1 and $P_3$ 80% token0 and 20% token1. Effectively this means that each position has the following internal ratio:
```
akure, line 166: @LaurensKubat  I think, an example with real tokens example with values would give better visualization. 
Like uosmo as token0 and ibc uatom as token 1, And deposit is 100000000uosmo, and 500000000 ibc atom, then what will be the internal ratio, external ratio, actual ratio and effective ratio.

smart-contracts/contracts/cl-vault/src/math/liquidity.rs
```diff
@@ -93,28 +148,28 @@ mod tests {
...
         let result = _liquidity1(amount1_desired, current_sqrt_p, sqrt_p_low).unwrap();
         assert_eq!(result.to_string(), "1517882343.751510418088349649");
     }
 
     #[test]
     fn test_max_liquidity0() {
-        let max_sqrt_price = Decimal::raw(10000000000000000000000000000000000000_u128);
-        let max_sqrt_price_low = Decimal::raw(300000000000000000000000000000000_u128);
-        let amount0_desired: Decimal = Decimal::from_ratio(1000000_u128, 1_u128);
+        let max_sqrt_price = Decimal256::raw(10000000000000000000000000000000000000_u128);
```
akure, line 161: what to assert in this max value overflow test?

smart-contracts/contracts/cl-vault/src/math/liquidity.rs
```diff
@@ -93,28 +148,28 @@ mod tests {
...
         // we only care about overflows here
         let _ = _liquidity0(amount0_desired, max_sqrt_price, max_sqrt_price_low).unwrap();
     }
 
     #[test]
     fn test_max_liquidity1() {
-        let max_sqrt_price = Decimal::raw(10000000000000000000000000000000000000_u128);
-        let max_sqrt_price_low = Decimal::raw(1000000000000000000000000000000000000_u128);
-        let amount0_desired: Decimal = Decimal::from_ratio(1000000_u128, 1_u128);
+        let max_sqrt_price = Decimal256::raw(10000000000000000000000000000000000000_u128);
```
akure, line 170: what to assert in this max value overflow test?

smart-contracts/contracts/cl-vault/src/msg.rs
```diff
@@ -40,24 +41,89 @@ pub enum AdminExtensionExecuteMsg {
...
+    IncreaseFunds {
+        position_id: u64,
+        token0: Coin,
+        token1: Coin,
+    },
+    DecreaseFunds {
+        position_id: u64,
+        liquidity: Decimal256,
+    },
+    /// Create a new position. This consumes all free balance up to max_percentage current free balance
```
njerschow, line 106: max_percentage no longer used? why not?

smart-contracts/contracts/cl-vault/src/helpers.rs
```diff
@@ -290,147 +326,201 @@ pub fn get_max_utilization_for_ratio(
...
+    _deps: DepsMut,
+    positions: Vec<(Position, FullPositionBreakdown)>,
+    spot_price: Decimal,
+    token0: Uint128,
+    token1: Uint128,
+) -> Result<Vec<(Position, Uint128, Uint128)>, ContractError> {
+    // get the wanted ratio per position
+    let ps = get_min_ratio_per_position(positions, spot_price)?;
+
+    let (total0, total1) = ps.iter().try_fold(
```
njerschow, line 340: is this step required if we normalize the ratio two lines before? total0 and total1 should be 1 right?

smart-contracts/contracts/cl-vault/src/helpers.rs
```diff
@@ -290,147 +326,201 @@ pub fn get_max_utilization_for_ratio(
...
+        },
+    )?;
+
+    // assuming we have the ratio wanted_token0/total_token0 and wanted_token1/total_token1
+    // ratio = min(wanted_token0/total_token0, wanted_token1/total_token1)
+    // for token0 and token1, we want to allocate
+    // allocate0 = token0 * ratio
+    // allocate1 = token1 * ratio
+    // afterwards we need to normalize all ratios to a 1 total
+    let ratios = ps.into_iter().map(|(ps, ps_ratio)| {
```
njerschow, line 353: major flashback to OPRO 😅

smart-contracts/contracts/cl-vault/src/helpers.rs
```diff
@@ -290,147 +326,201 @@ pub fn get_max_utilization_for_ratio(
...
+        let r0 = Decimal::from_ratio(ps_ratio.asset0, total0);
+        let r1 = Decimal::from_ratio(ps_ratio.asset1, total1);
+        let ratio = min(r0, r1);
+
+        (ps, ratio)
+    });
+
+    let total_ratio = ratios
```
njerschow, line 361: minor nit but we could be using a struct for this to keep things normed

njerschow:
@LaurensKubat can we get a fast-forward or rebase from main here?

PR #564, opened 2024-01-22
Arhamchordia/auto compounding

arhamchordia:
## 4. Checklist

### DIFF
--- a/smart-contracts/contracts/cl-vault/src/rewards/distribution.rs
+++ b/smart-contracts/contracts/cl-vault/src/rewards/distribution.rs
@@ -1,5 +1,8 @@
 use apollo_cw_asset::AssetInfo;
-use cosmwasm_std::{Addr, Coin, CosmosMsg, Deps, DepsMut, Env, Event, Response, StdError, SubMsg, SubMsgResult, to_json_binary, Uint128, WasmMsg};
+use cosmwasm_std::{
+    to_json_binary, Addr, Coin, CosmosMsg, Deps, DepsMut, Env, Event, Response, StdError, SubMsg,
+    SubMsgResult, Uint128, WasmMsg,
+};
 use cw_dex_router::{
     msg::{BestPathForPairResponse, ExecuteMsg, QueryMsg},
     operations::SwapOperationsListUnchecked,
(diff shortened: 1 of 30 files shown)


smart-contracts/contracts/cl-vault/src/rewards/distribution.rs
```diff
@@ -159,101 +81,188 @@ pub fn handle_collect_spread_rewards_reply(
...
-            // Merge the new reward with any existing rewards for the user
-            USER_REWARDS.update(
-                deps.storage,
-                user_address,
-                |old| -> ContractResult<CoinList> {
-                    Ok(match old {
-                        Some(old_rewards) => user_reward_clone.add(old_rewards)?,
-                        None => user_reward_clone,
-                    })
+    let current_swap_route = swap_routes[0].clone();
```
magiodev, line 117: Probably here we want to enhance this handling. Looks like we miss a logic step which is detecting if any given route (either 1 or more) is related to any denom not included in the pool pair. So we want to swap those first i.e. N% N% to asset0 and asset1 or 100% asset0 or asset1 based on swap efficiency (we will have to rearrange percentages on the next re-range anyway so swap into a single asset is likely to be simpler).

smart-contracts/contracts/cl-vault/src/rewards/distribution.rs
```diff
@@ -159,101 +81,188 @@ pub fn handle_collect_spread_rewards_reply(
...
+    env: Env,
+    force_swap_route: bool,
+    mut swap_routes: Vec<AutoCompoundAsset>,
 ) -> Result<Response, ContractError> {
-    let mut rewards_status = REWARDS_STATUS.load(deps.storage).unwrap();
-    if rewards_status != RewardsStatus::Distributing {
-        return Err(ContractError::IsNotDistributing {});
+    // auto compound admin
+    let dex_router = DEX_ROUTER.may_load(deps.storage)?;
+    if swap_routes.is_empty() {
```
magiodev, line 113: Are we sure we want to error if any route is provided? Should the cw-dex-router being able to provide one by default? At this point from a design standpoint is like this entrypoint is a "swap_external_to_pool_denom", which prepares the vault for autocompounding idle assets and rewards, but not directly related to the autocompound action.

LaurensKubat (commented):
This version is still permissioned, which is fine for now as long as we make proper note of our intention to make that permissionless. Both in and outside of our code

smart-contracts/contracts/cl-vault/src/rewards/distribution.rs
```diff
@@ -159,101 +81,188 @@ pub fn handle_collect_spread_rewards_reply(
...
-            |mut old_rewards| -> ContractResult<CoinList> {
-                old_rewards.sub(&distributed_rewards)?;
-                Ok(old_rewards)
-            },
-        )?;
-        // Clear distributed rewards state
-        DISTRIBUTED_REWARDS.save(deps.storage, &CoinList::new())?;
-    } else {
-        DISTRIBUTED_REWARDS.save(deps.storage, &distributed_rewards)?;
+        response.clone().add_message(next_autocompound_msg);
```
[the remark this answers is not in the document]
arhamchordia: For this, Laurens suggested to stack all of them together for swapping instead of one by one.

smart-contracts/contracts/cl-vault/src/rewards/distribution.rs
arhamchordia: Instead of this we are pool them all together as sub messages array.

arhamchordia:
This PR changes have been added to https://github.com/quasar-finance/quasar/pull/586/

PR #565, opened 2024-01-29
`migration`: inverted pool

ajansari95:
## 1. Overview
- adds changes to migrate to pool with inverted asset composition

## 2. Implementation details
- Implementation same as earlier 
- Change is that the migrate msg takes the new ticks and vault assets are inverted

## 3. How to test/use
- tested via a migration on testnet

## 4. Checklist


magiodev (approved):
Approach looks good to me, which is to pass from the message the new lower and upper tick as changing both, pool and asset order, would invalidate the previous position's ticks. And other changes accordingly. Approved for me.

PR #575, opened 2024-02-28
add test where we create merkle root during test run

njerschow:
## 4. Checklist


magiodev (approved):
lgtm! i just removed a couple of println! i forgot there from previous debuggings

PR #579, opened 2024-03-06
Update get_min_ratio_per_position math logic

njerschow:
## 1. Overview

Improving the logic within the get_min_ratio_per_position function to remove the dependency on an arbitrary liquidity amount. Algebra in README

## 2. Implementation details

After deriving the equation, we simply needed to rewrite the math according to the new formulas

## 3. How to test/use

cargo test-tube

## 4. Checklist


LaurensKubat (commented):
Looks interesting to me, would love to see the formulas work out and a small implementation. It would def make for better and cleaner code

LaurensKubat (commented):
the code lgtm, but some tests with big and and small potential numbers would be a good add

PR #583, opened 2024-03-19
Vault incentives

magiodev:
## 4. Checklist

### DIFF
--- a/smart-contracts/contracts/merkle-incentives/src/incentives/mod.rs
+++ b/smart-contracts/contracts/merkle-incentives/src/incentives/mod.rs
@@ -84,44 +84,46 @@ impl From<Vec<Coin>> for CoinVec {
 }
 
 impl PartialOrd for CoinVec {
-    // todo: test this
     fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
-        // if even one in self is greater than other, this should return greater
-        for coin in &self.0 {
-            if let Some(other_coin) = other.0.iter().find(|c| c.denom == coin.denom) {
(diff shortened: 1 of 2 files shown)


smart-contracts/contracts/merkle-incentives/src/incentives/mod.rs
```diff
@@ -84,44 +84,60 @@ impl From<Vec<Coin>> for CoinVec {
...
+    // - If `self_less` is false and `self_greater` is true, it returns `Greater`.
+    // - If both flags are false, it returns `Equal`.
+    // - If both flags are true, it returns `None`, indicating incomparability
     fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
-        // if even one in self is greater than other, this should return greater
-        for coin in &self.0 {
-            if let Some(other_coin) = other.0.iter().find(|c| c.denom == coin.denom) {
-                if coin.amount > other_coin.amount {
-                    return Some(std::cmp::Ordering::Greater);
+        let self_map: std::collections::HashMap<_, _> = self
```
LaurensKubat, line 102: Theoretically, iterating through lists with less than a couple thousand items is faster than a hashmap. However that does not matter much right now with gas prices. No action needed but we should use `Vec` in the package

LaurensKubat (approved):
lgtm

smart-contracts/contracts/merkle-incentives/src/incentives/mod.rs
arhamchordia: Added todo for later

PR #581, opened 2024-03-13
Smart contracts workspace improvements

magiodev:
## 1. Overview

In this pull request, we've made several enhancements and refinements to the smart contracts workspace to improve its overall functionality and maintainability.

## 2. Implementation details

- **Code Quality Improvements:**
  - Resolved various warnings to enhance the code quality and stability of the workspace (Commit: e7c0020).
  - Polished existing tests to ensure their reliability and coverage (Commit: 1610f31).

- **Build System Updates:**
  - Introduced `justfile` to streamline the build process (Commit: b93839b).

- **Codebase Modernization:**
  - Updated serialization methods: Replaced `to_binary` with `to_json_binary` and `from_json` with `from_json_binary` to align with current standards (Commit: 639b4e3).
  - Fixed and replaced deprecated serialization functions to enhance maintainability (Commits: fd2f635, db5eae6).

- **Dependency Management:**
  - Added more dependencies to the workspace to ensure consistency and ease of management (Commit: a588eb2).
  - Updated `prost` dependencies across all contracts and included them in the workspace configuration to maintain compatibility and performance (Commits: 3f10156, e7eb9b1, 22e73ca).
  - Bumped additional dependencies to their latest versions and incorporated them into the workspace configuration for improved functionality (Commit: 535f190).

- **CI/CD and Build Script Adjustments:**
  - Removed the make command, updated CI to use the just command, and eliminated unused root directories to streamline the build and deployment process (Commit: 80a2e04).

- **Code Reversions and Fixes:**
  - Conducted further polishing and reverted specific changes in the `cl-vault` related to the autocompound feature to align with the intended functionality (Commits: 5614754, c60d73b).

- **Miscellaneous:**
  - Merged the latest changes from the main branch to ensure that the workspace is up-to-date with the base code (Commit: 7d62a7d).
  - A final note indicating a successful return to a stable state (Commit: c60d73b).

## 4. Checklist

- [ ] Does the Readme need to be updated?

PR #585, opened 2024-03-22
Adding burn contract.

arhamchordia:
## 1. Overview

In this PR we are adding Burn contract that will burn the token sent to it.

## 2. Implementation details

The Burn Coins Contract is a simple smart contract designed to burn the coins sent to it. Here are some highlights of the implementation:

- The contract consists of an `instantiate` function to initialize the contract.
- It includes an `execute` function that has `burn` function to handle the execution of coin burning.
- The contract allows to query the total amount of coins burnt using the `query` function.

Design Rationale:
- The contract follows a straightforward design to achieve its primary objective of burning coins.
- It ensures that burning coins is irreversible to maintain the integrity of the tokenomics.

## 3. How to test/use

To test or use the Burn Coins Contract, follow these steps:
- Build the contract locally
- Change to demos directory
- Run `sh quasar_localnet.sh`
- Run `burn.sh`

## 4. Checklist

## 6. Future Work (optional)

Potential future enhancements for the Burn Coins Contract include:
- Implementing additional functionality for managing burnt coins, such as tracking burnt amounts per user or token.
- Adding support for configurable burning parameters, such as burn fees or burn rates.


smart-contracts/contracts/burn_coins/src/state.rs
```diff
@@ -0,0 +1,30 @@
+use cosmwasm_std::Coin;
+use cw_storage_plus::Item;
+
+pub const AMOUNT_BURNT: Item<Vec<Coin>> = Item::new("burnt_coins");
```
LaurensKubat, line 4: Since total burnt here is an item and a `Vec` could theoretically contain multiple coins, meaning that the entire Vec would have to be loaded for us to update one coin. Cleaner would be to use a `Map`, with the denom as key and the value as the amount

LaurensKubat (changes requested):
state should be a bit cleaner, but general structure looks good

smart-contracts/contracts/burn_coins/src/query.rs
```diff
@@ -0,0 +1,103 @@
+use cosmwasm_std::{Coin, Deps, Order, StdResult};
+
+use crate::msg::TotalBurntResponse;
+use crate::state::AMOUNT_BURNT;
+
+pub fn query_total_burn(deps: Deps) -> StdResult<TotalBurntResponse> {
+    let mut total_burn_amount: Vec<Coin> = vec![];
+
+    for res in AMOUNT_BURNT.range(deps.storage, None, None, Order::Ascending) {
```
LaurensKubat, line 9: nitpick: this can be a bit cleaner

this can also be done by collecting the range in a vec, something like:
```
let total_burn_amount: Result<Vec<Coin>, ContractError> = AMOUNT_BURNT.range(deps.storage, None, None, Order::Ascending).collect
```
Depending on the contents of AMOUNT_BURNT, you might need to do some more work through `.map()`, such as wrapping the key and value in a coin

LaurensKubat (approved):
lgtm, one nitpick comment

smart-contracts/contracts/burn_coins/src/contract.rs
```diff
@@ -0,0 +1,140 @@
...
+    if info.funds.is_empty() {
+        return Err(BurnErrors::ZeroAmount {});
+    }
+
+    // Prepare the Burn message
+    let burn_msg = CosmosMsg::Bank(BankMsg::Burn {
+        amount: info.clone().funds,
+    });
+
+    for fund in info.funds {
```
magiodev, line 65: This could be a bit cleaner like

```
for fund in &info.funds {
    let denom = &fund.denom;
    let amount = &fund.amount;
    let mut total_burn_amount = AMOUNT_BURNT.may_load(deps.storage, denom)?.unwrap_or_default();
    total_burn_amount += amount;
    AMOUNT_BURNT.save(deps.storage, denom, &total_burn_amount)?;
}

```

smart-contracts/Cargo.toml
```diff
@@ -4,6 +4,7 @@ members = [
   "contracts/airdrop",
   "contracts/lp-strategy",
   "contracts/basic-vault",
+  "contracts/burn_coins",
   "contracts/vault-rewards",
   "contracts/multihop-router",
   "contracts/cl-vault",
```
magiodev, line 10: Dependencies should inherit from workspace now that PR #581 has been merged

magiodev (changes requested):

smart-contracts/contracts/burn_coins/src/state.rs
```diff
@@ -0,0 +1,58 @@
+use cosmwasm_std::Uint128;
+use cw_storage_plus::Map;
+
+pub const AMOUNT_BURNT: Map<String, Uint128> = Map::new("burnt_coins");
```
magiodev, line 4: Is that needed? What is the purpose of keeping track of burned coins if we can simply `query bank balances` over the contract's addy?

magiodev (commented):
just an open discussion to KISS

smart-contracts/contracts/burn_coins/src/state.rs
LaurensKubat: coins would not be in the contract's addy since we call the bank modules burn

smart-contracts/contracts/burn_coins/src/contract.rs
```diff
@@ -0,0 +1,140 @@
...
+        QueryMsg::TotalBurntQuery {} => to_json_binary(&query_total_burn(deps)?),
+    }
+}
+
+#[entry_point]
+pub fn migrate(_deps: DepsMut, _env: Env, _msg: MigrateMsg) -> Result<Response, BurnErrors> {
+    Ok(Response::new().add_attribute("migrate", "successful"))
+}
+
+pub fn execute_coins_burn(deps: DepsMut, info: MessageInfo) -> Result<Response, BurnErrors> {
```
magiodev, line 55: Just an autistic comment: `execute_burn` looks a better name to me lol

smart-contracts/contracts/burn_coins/src/contract.rs
```diff
@@ -0,0 +1,140 @@
...
+        {
+            total_burn_amount += fund.clone().amount;
+            AMOUNT_BURNT.save(deps.storage, fund.clone().denom, &total_burn_amount)?;
+        } else {
+            AMOUNT_BURNT.save(deps.storage, fund.clone().denom, &fund.clone().amount)?;
+        }
+    }
+
+    // Return a response with the burn message
+    Ok(Response::default().add_message(burn_msg))
```
magiodev, line 77: This looks indeed correct. But im just wondering in the case we want to build a Subgraph indexer, would we need some extra attribute here? If any it should be something very minimalistic, but if too much minimalistic it could be useless. So this could be probably skipped in favor of an hardcoded contract addy on the subgraph side, looking for sdk.Burn events.

What do you think?

smart-contracts/contracts/burn_coins/src/contract.rs
LaurensKubat: makes sense

smart-contracts/contracts/burn_coins/src/contract.rs
arhamchordia: This can be indexed by checking the bank module burn event. That will be a more reliable event to index.

smart-contracts/contracts/token-burner/src/contract.rs
```diff
@@ -0,0 +1,139 @@
...
+    if info.funds.is_empty() {
+        return Err(BurnErrors::ZeroAmount {});
+    }
+
+    // Prepare the Burn message
+    let burn_msg = CosmosMsg::Bank(BankMsg::Burn {
+        amount: info.clone().funds,
+    });
+
+    for fund in &info.funds {
```
LaurensKubat, line 65: nit: I think we could remove the clones on denom if we remove the & on `let denom = &fund.denom` and instead do fund.denom

LaurensKubat (commented):
2 small comments that I missed first time around, otherwise LGTM

LaurensKubat (approved):
lgtm

PR #586, opened 2024-03-28
Feat/deposit math change

arhamchordia:
## 1. Overview

- Change ExactDeposit to no longer add to position, but just deposit on ratio and refund the rest
- Add AnyDeposit, that swaps internally
- Add Autocompounding

### Exact Deposit
exact deposit no longer adds funds to the vaults current position, but instead adds the funds to the vaults free balance. In doing so, the vaults are more able to handle the dynamic Osmosis uptimes. This change should also result in smaller refund amounts. Using the autocompounding endpoints, these funds can later be added to the vaults position

### Any Deposit
This PR adds an endpoint for users to deposit any amount of token0 or token1 into the vault. The vault then swaps these tokens in the needed ratio. The swapped tokens then become part of the vaults free balance. Using the autocompounding endpoints, these funds can later be added to the vaults position

### Autocompounding
This PR adds multiple endpoints to the CL for autocompounding. The three endpoints are:
1. CollectRewards, collects the accrued rewards from the position and saves the amount
2. AutoCompoundRewards, this swaps the rewards to the current ratio of the vault, preparing them for redepositing
3. Redeposit, this adds all funds to the position without doing any further swapping
The implementations of exact depsoit and any deposit depend on redeposit being present for adding funds to the vault

## 3. How to test/use

Each different feature can be tested fully in test tube. Autocompounding can be fully tested separate from the deposit methods. The deposit methods should include a call to redeposit for good measure.
For good measure, the rollout of this PR should be done in steps to make sure all off chain infrastructure is properly adapted

## 4. Checklist

Yes

## 5. Limitations (optional)

- The current AnyDeposit does not allow for swap routes to be added to the internal swap.
- Both deposit methods do not fully add a concept of uptimes

## 6. Future Work (optional)

A separate AnyDeposit method without a recipient field should added for Authz integrations such as Yieldmos


ajansari95 (changes requested):
ci failing for unit tests and lint. Also looks like build is failing

smart-contracts/contracts/cl-vault/src/msg.rs
```diff
@@ -1,32 +1,57 @@
...
     query::{
         AssetsBalanceResponse, PoolResponse, PositionResponse, RangeAdminResponse,
-        UserRewardsResponse, UserSharesBalanceResponse, VerifyTickCacheResponse,
+        UserSharesBalanceResponse, VerifyTickCacheResponse,
     },
     state::{Metadata, VaultConfig},
 };
 
+#[cw_serde]
+pub struct SwapAsset {
```
LaurensKubat, line 15: I think this should not be in a state file but in an autocompounding related file

smart-contracts/contracts/cl-vault/src/rewards/distribution.rs
LaurensKubat, line 1: This needs to handle the edge cases of 100% and 0% strategist fees

smart-contracts/contracts/cl-vault/src/test_tube/authz.rs
LaurensKubat, line 1: If we have time, would be nice to add a test asserting that any_deposit authz is the same as regular any deposit

smart-contracts/contracts/cl-vault/src/vault/any_deposit.rs
```diff
@@ -0,0 +1,337 @@
...
+        coins_to_mint_for.push(Coin {
+            denom: pool_config.token1.clone(),
+            amount: balance1 + deposit_amount_in_ratio.1,
+        });
+    }
+
+    let (mint_msg, user_shares) = mint_msg_user_shares(
+        deps.branch(),
+        &env,
+        &(coins_to_mint_for[0].amount, coins_to_mint_for[1].amount),
```
LaurensKubat, line 189: Would this not fail for single sided positions?, if we don't hit one of the above if cases, we only send an array of length 1. Better would be to work with a tuple here, pushing to an array also gives us the risk of mistaking token0 for token1 and vice versa

LaurensKubat (changes requested):
Some small comments/changes, Also should this PR come with a range middleware update for the claim_after time?

smart-contracts/contracts/cl-vault/src/msg.rs
magiodev: moved

smart-contracts/contracts/cl-vault/src/test_tube/authz.rs
magiodev: added

smart-contracts/contracts/cl-vault/src/rewards/distribution.rs
magiodev: This was handled inside the helpers::CoinList mul_ratio, fixed in this PR. I added tests to assert those cases tho.

LaurensKubat (approved):
lgtm

LaurensKubat (approved):
my own new changes also lgtm
download

smart-contracts/contracts/cl-vault/src/msg.rs
```diff
@@ -1,32 +1,51 @@
...
-    CollectRewards { amount_of_users: Uint128 },
+    /// provides an entry point for autocompounding idle funds to current position
+    Autocompound {},
     /// Distribute any rewards over all users
-    DistributeRewards { amount_of_users: Uint128 },
-    /// Claim rewards belonging to a single user
-    ClaimRewards {},
+    CollectRewards {},
+    /// Build tick exponent cache
+    BuildTickCache {},
```
magiodev, line 32: This entrypoint is exposed twice, here and in vault::admin.rs as `AdminExtensionExecuteMsg`, we should remove one.

smart-contracts/contracts/cl-vault/src/msg.rs
```diff
@@ -141,6 +159,8 @@ pub struct InstantiateMsg {
     pub admin: String,
     /// Address that is allowed to update range.
     pub range_admin: String,
+    /// Address that is allowed to auto compound
+    pub auto_compound_admin: String,
```
magiodev, line 163: I think this is unused now, we only rely on range_admin

smart-contracts/contracts/cl-vault/src/state.rs
```diff
@@ -28,21 +27,42 @@ pub struct VaultConfig {
     pub performance_fee: Decimal,
     /// Account to receive fee payments
     pub treasury: Addr,
-    /// swap max slippage
+    /// swap max slippage // TODO: This is unused
     pub swap_max_slippage: Decimal,
```
magiodev, line 31: This is also unsued

magiodev (commented):
Added some minor comments, but lgtm!

ajansari95 (dismissed):
lgtm 🚢

LaurensKubat (approved):
lgtm

PR #587, opened 2024-04-01
Feat/deposit math change incentives

magiodev:
## 1. Overview

This PR introduces a significant change in the way incentives are claimed within our system. We've adjusted the claim order, now allowing incentives to be claimed optionally as the second step. Key to this update is the introduction of a `claim_after` parameter in `ModifyRangeMsg`, which dictates a time threshold that must pass before incentives can be claimed. This addition aims to prevent the premature forfeiture of incentives.

Additionally, we've enriched the `POSITION` state with a `join_time` attribute, derived from `env.block.time.seconds()`. While it's possible to extract `join_time` via an additional query step to `position.join_time`, this current implementation sidesteps unnecessary complexity without sacrificing functionality.

## 2. Implementation details

- Modified the claim process order: Incentives can now be claimed after spread rewards.
- Added `claim_after` to `ModifyRangeMsg`, applying it to the `POSITION` state.
- Integrated `join_time` within `POSITION`, sourced from `env.block.time.seconds()`.

**Benefits:**
- Ensures that incentives are claimed in a timely manner, reducing the chance of forfeiture.

## 4. Checklist


LaurensKubat (commented):
some small comments but generally lgtm

LaurensKubat (approved):
lgtm

PR #592, opened 2024-04-17
fix: revert swap reply logic in cl-vault

ajansari95:
## 4. Checklist


magiodev (approved):
revert of commit lgtm. we'll address that in the next patch

LaurensKubat (approved):
lgtm

PR #595, opened 2024-04-25
fix: migrate entrypoints in `cl-vault` and `range-middleware`

ajansari95:
## 1. Overview
- adds migrate entrypoint in `range-middleware`
- fixes the migration entrypoint in `cl-vault`

## 2. Implementation details
- The `range-middleware` contract was missing migrate entry point this PR adds an empty migrate entrypoint with subsequent `MigrateMsg`
- A migration to initialise the paginated distribution states is added in `cl-vault`

## 4. Checklist


LaurensKubat (approved):
lgtm

PR #597, opened 2024-05-02
Feat/add claiming before actions

LaurensKubat:
## 1. Overview

Claim rewards before doing any actions that move the position

## 2. Implementation details

Adds a function that adds a callback to the contracts own claim message before any other messages to a response. 

## 3. How to test/use

check that the first message of redeposit and move_range is a claim msg of the vault

## 4. Checklist


magiodev (approved):
looks good!

PR #599, opened 2024-05-13
Feat/fee splitter

LaurensKubat:
## 1. Overview

Adds a contract that splits fees from the contract balance and sends those to the preset receivers

## 2. Implementation details

Adds a contract that supports claiming as well as receiving funds via bank transfers.
The claiming implementation works by calling a contract with an arbitrary message, thus this contract would be extremely unsafe for any CW20 usage, but would well for native tokens since we can force the tokens sent to 0 for native tokens

## 3. How to test/use

This contract has very little cross contract test, so almost all functionality can be tested in either a unit test or a simple multi test

## 4. Checklist


arhamchordia (approved):
Looks good. Do we keep the claim feature enable by default?

LaurensKubat:
> Looks good. Do we keep the claim feature enable by default?
 Depends on how we want to deploy, if we write all strategists claims with bank sends, that shouldn't be necessary. The reason I made is optional is because in general, reducing complexity surface is a good thing. So If we can not do so, I would strongly prefer to not do so

PR #602, opened 2024-05-20
Feat deposit math change authz tests

LaurensKubat:
## 1. Overview

Adds test asserting same behaviour of authz endpoint


magiodev (approved):
looks good!

PR #604, opened 2024-05-21
Adding code and test for removing icq which have been retried.

arhamchordia:
## 1. Overview

Adds code for removing already parsed ICQs

## 3. How to test/use

Test in the same file.

### DIFF
--- a/smart-contracts/contracts/lp-strategy/src/contract.rs
+++ b/smart-contracts/contracts/lp-strategy/src/contract.rs
@@ -1,8 +1,8 @@
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
 use cosmwasm_std::{
-    from_json, Attribute, DepsMut, Env, IbcMsg, IbcPacketAckMsg, MessageInfo, QuerierWrapper,
-    Reply, Response, Storage, Uint128,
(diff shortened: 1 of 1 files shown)


smart-contracts/contracts/lp-strategy/src/contract.rs
```diff
@@ -271,6 +273,20 @@ pub fn execute_try_icq(deps: DepsMut, env: Env) -> Result<Response, ContractErro
     } else {
         res = res.add_attribute("IBC_LOCK", "locked");
     }
+
+    #[allow(clippy::type_complexity)]
+    let traps: Result<Vec<((u64, String), Trap)>, ContractError> = TRAPS
```
LaurensKubat, line 278: This would out gas and potentially brick the contract here

LaurensKubat (changes requested):
Instead of doing this in the try_icq entrypoint where it would intervene in regular logic, this match in execute_retry should be extended: ```match traps.step {
        IbcMsgKind::Ica(ica_kind) => match ica_kind {
            IcaMessages::ExitPool(pending) => {
                handle_retry_exit_pool(deps, env, pending, seq, channel)
            }
            IcaMessages::JoinSwapExternAmountIn(pending) => {
                handle_retry_join_pool(deps, env, pending, seq, channel)
            }
            _ => todo!(),
        },
        _ => todo!(),
    }```

PR #616, opened 2024-06-06
swap logic generalize

magiodev:
## 4. Checklist

### DIFF
--- a/smart-contracts/contracts/cl-vault/src/vault/any_deposit.rs
+++ b/smart-contracts/contracts/cl-vault/src/vault/any_deposit.rs
@@ -1,7 +1,7 @@
 use apollo_cw_asset::AssetInfoBase;
 use cosmwasm_std::{
-    coin, Addr, Coin, CosmosMsg, Decimal, DepsMut, Env, Fraction, MessageInfo, Response, SubMsg,
-    SubMsgResult, Uint128, Uint256,
+    coin, Addr, Coin, Decimal, DepsMut, Env, Fraction, MessageInfo, Response, SubMsg, SubMsgResult,
+    Uint128, Uint256,
 };
 use cw_dex::Pool::Osmosis;
 use cw_dex_router::operations::{SwapOperationBase, SwapOperationsListUnchecked};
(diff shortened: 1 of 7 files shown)


smart-contracts/contracts/cl-vault/src/vault/any_deposit.rs
```diff
@@ -84,13 +83,21 @@ pub fn execute_any_deposit(
         // rest minting logic remains same
         Ok(Response::new()
             .add_submessage(SubMsg::reply_on_success(
-                swap_msg,
+                swap_calc_result.swap_msg,
```
lubkoll, line 86: We create the same message below in line 129. We could just read the data for the attributes in the if-else-branch and then have one return statement instead of one for each branch

smart-contracts/contracts/cl-vault/src/vault/any_deposit.rs
```diff
@@ -308,28 +322,35 @@ fn swap_msg_token_in_out_amounts(
     let token_out_min_amount = token_out_ideal_amount?
         .checked_multiply_ratio(max_slippage.numerator(), max_slippage.denominator())?;
 
+    if !pool_config.pool_contains_token(token_in_denom) {
+        return Err(ContractError::InvalidSwapAssets {});
+        // TODO: Restore this error
```
lubkoll, line 327: What's the issue that we don't have this anymore, but need it back?

smart-contracts/contracts/cl-vault/src/helpers/msgs.rs
```diff
@@ -0,0 +1,175 @@
...
+            &params.token_in_denom.to_string(),
+            params.token_out_min_amount,
+        ));
+    }
+
+    // we know we have a dex_router, so we can unwrap
+    let dex_router_address = dex_router.clone().unwrap();
+
+    // if we need to force the route
+    if params.forced_swap_route.is_some() {
```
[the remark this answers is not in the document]
magiodev: Good catch, fixed!

smart-contracts/contracts/cl-vault/src/vault/any_deposit.rs
magiodev: put back. also on range.rs

lubkoll (approved):

PR #618, opened 2024-06-18
Update build.yml

arhamchordia:
Removing contract build for arm64.

## 4. Checklist


magiodev (approved):
looks good! i agree this can be avoided

PR #620, opened 2024-06-24
Feat/dex router

lubkoll:
Current state:
- swaps work
- cw20 removed
- changed computation of best path to use single query
- on-chain computation of optimal swap route
- Rename to dex-router-osmosis


smart-contracts/Cargo.toml
```diff
@@ -32,6 +33,8 @@ cw20-base = {version = "1.1.2", features = ["library"]}
 cw20-staking = "0.11.1"
 apollo-cw-asset = "0.1.2"
 cw-dex = {git = "https://github.com/quasar-finance/cw-dex", branch = "feat/deprecate-osmo-gamm"}
+mars-owner = "2.0.0"
```
magiodev, line 36: I dont know if we should avoid add this dependency only to use the NotOwner error. Can we replace it for a local Unauthorized {} enum variant on the error.rs?

smart-contracts/contracts/dex-router-osmosis/src/msg.rs
```diff
@@ -0,0 +1,68 @@
...
+        offer_denom: String,
+        ask_denom: String,
+        path: Vec<u64>,
+        bidirectional: bool,
+    },
+    RemovePath {
+        offer_denom: String,
+        ask_denom: String,
+        path: Vec<u64>,
+        bidirectional: bool,
```
magiodev, line 26: Can this field be removed? I think we only need the index selectors here during the remove.

smart-contracts/contracts/dex-router-osmosis/src/state.rs
```diff
@@ -0,0 +1,15 @@
...
+use mars_owner::Owner;
+use osmosis_std::types::osmosis::poolmanager::v1beta1::SwapAmountInRoute;
+
+#[cw_serde]
+pub struct RecipientInfo {
+    pub address: Addr,
+    pub denom: String,
+}
+
+pub const PATHS: Map<(String, String), Vec<Vec<SwapAmountInRoute>>> = Map::new("paths");
```
magiodev, line 13: Can we comment at line 12 what the index String, String represents?

magiodev (changes requested):
Mostly looks good! Minor changes requested.

smart-contracts/contracts/dex-router-osmosis/src/state.rs
lubkoll: done

smart-contracts/contracts/dex-router-osmosis/src/msg.rs
lubkoll: Yes. I have it here for symmetry. Both in SetPath and RemovePath we could leave it out. As it is in SetPath, the interface looks a bit more consistent to me with having it here too. I can remove it if you prefer to, else I will keep it.

smart-contracts/Cargo.toml
lubkoll: As these checks for the sender are the same in all contracts it is quite nice to reuse existing solutions for that. The mars-owner is just the new version of the Admin in cw-controllers. It has a better implementation and -- though not used here -- a quite safe mechanism for transfer of ownership.

smart-contracts/contracts/dex-router-osmosis/README.md
```diff
@@ -0,0 +1,9 @@
+# Dex router for osmosis
+
+Responsible for the on-chain computation of multi-hop swap routes.
+Currently only supports pools with 2 tokens.
+
+TODO:
+ * Fix queries for best path
+ * Add checks and unit-tests
+ * Add support for (optional) on-chain detection of the optimal swap path
```
arhamchordia, line 9: Can add a todo for test-tube tests.

smart-contracts/Cargo.toml
magiodev: I was going to suggest to use cw-controllers but ok, sounds good!

smart-contracts/contracts/dex-router-osmosis/README.md
lubkoll: Thx, actually that list wasn't updated properly by me. Fixed it.
Now, the only item is "possibly test-tube tests". Wrote possibly as I have some concerns that that might be more work than we get in return from it.

smart-contracts/contracts/dex-router-osmosis/src/msg.rs
magiodev: If it is useful we can keep it. If not we should remove

smart-contracts/contracts/dex-router-osmosis/src/contract.rs
```diff
@@ -0,0 +1,435 @@
...
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn instantiate(
+    deps: DepsMut,
+    _env: Env,
+    info: MessageInfo,
+    _msg: InstantiateMsg,
+) -> Result<Response, ContractError> {
+    set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+    OWNER.initialize(
```
ajansari95, line 34: here `OWNER` -(admin) is set on the sender address, which cant be updated afterwards. We should have a `updatOwner` or `updateAfmin` endpoint. @lubkoll @magiodev ?

lubkoll: done

ajansari95 (dismissed):
lgtm

magiodev (dismissed):
looks good!

magiodev (approved):

ajansari95 (approved):
lgtm

arhamchordia (approved):
LGTM

PR #638, opened 2024-06-28
add tests for `dex-router-osmosis`

ajansari95:
adds test for working flow in dex router for
1. set path
2. remove path
3. swap

### DIFF
--- a/smart-contracts/contracts/dex-router-osmosis/src/tests/initialize.rs
+++ b/smart-contracts/contracts/dex-router-osmosis/src/tests/initialize.rs
@@ -1,0 +1,482 @@
+use std::str::FromStr;
+
+use cosmwasm_schema::cw_serde;
+use cosmwasm_std::{Addr, Attribute, Coin, Decimal, Uint128};
+use osmosis_std::types::cosmos::base::v1beta1::Coin as OsmoCoin;
+use osmosis_std::types::osmosis::concentratedliquidity::v1beta1::{
+    CreateConcentratedLiquidityPoolsProposal, Pool, PoolRecord, PoolsRequest,
+};
+use osmosis_std::types::osmosis::gamm::v1beta1::MsgJoinPool;
+use osmosis_std::types::osmosis::poolmanager::v1beta1::SpotPriceRequest;
+use osmosis_std::types::osmosis::tokenfactory::v1beta1::{MsgCreateDenom, MsgMint};
+
+use osmosis_test_tube::osmosis_std::types::osmosis::gamm::poolmodels::balancer::v1beta1::MsgCreateBalancerPool;
+use osmosis_test_tube::osmosis_std::types::osmosis::gamm::v1beta1::PoolAsset;
+use osmosis_test_tube::Gamm;
+use osmosis_test_tube::{
+    cosmrs::proto::traits::Message,
+    osmosis_std::types::osmosis::concentratedliquidity::{
+        poolmodel::concentrated::v1beta1::MsgCreateConcentratedPool, v1beta1::MsgCreatePosition,
+    },
+    Account, ConcentratedLiquidity, GovWithAppAccess, Module, OsmosisTestApp, PoolManager,
+    SigningAccount, TokenFactory, Wasm,
+};
+
+use crate::msg::{BestPathForPairResponse, ExecuteMsg, InstantiateMsg, QueryMsg};
+
+pub(crate) const ADMIN_BALANCE_AMOUNT: u128 = 3402823669209384634633746074317682114u128;
+pub(crate) const TOKENS_PROVIDED_AMOUNT: &str = "1000000000000";
+pub(crate) const FEE_DENOM: &str = "uosmo";
+pub(crate) const DENOM_BASE: &str = "udydx";
+pub(crate) const DENOM_QUOTE: &str = "uryeth";
+pub(crate) const INTERMEDIATE_BASE: &str = "uinshallah";
+pub(crate) const INTERMEDIATE_QUOTE: &str = "wosmo";
+
+#[cw_serde]
+pub struct PoolWithDenoms {
+    pub pool: u64,
+    pub denom0: String,
+    pub denom1: String,
+}
+
+pub fn default_init() -> (OsmosisTestApp, Addr, Vec<PoolWithDenoms>, SigningAccount) {
+    let app = OsmosisTestApp::new();
+    let tf = TokenFactory::new(&app);
+
+    // Create new account with initial funds
+    let admin = app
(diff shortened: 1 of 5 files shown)


smart-contracts/contracts/dex-router-osmosis/src/tests/mod.rs
```diff
@@ -1,3 +1,4 @@
+mod initialize;
```
lubkoll, line 1: Could we have the test-tube tests in a separate directory or with some tag such that we can run them independently of the unit-tests?

ajansari95: i thing adding a tag makes sense as we have grouped test by the functions?

smart-contracts/contracts/dex-router-osmosis/src/tests/swap.rs
```diff
@@ -1,10 +1,15 @@
 use crate::contract::{execute, instantiate};
 use crate::error::ContractError;
-use crate::msg::{ExecuteMsg, InstantiateMsg};
-use cosmwasm_std::coin;
+use crate::msg::{ExecuteMsg, InstantiateMsg, QueryMsg};
 use cosmwasm_std::testing::{mock_dependencies, mock_env, mock_info};
+use cosmwasm_std::{coin, Coin, Uint128};
+use osmosis_std::types::osmosis::poolmanager::v1beta1::SwapAmountInRoute;
+use osmosis_test_tube::cosmrs::proto::tendermint::v0_37::abci::response;
```
lubkoll, line 7: can this be removed, seems to be unused?

magiodev:
Tests looks good, but I see redundancies with the ones we have on CL Vault contract. Maybe we should maintain all of them in one place.

Or if I am missing something, how are they different from the others?

ajansari95:
> Tests looks good, but I see redundancies with the ones we have on CL Vault contract. Maybe we should maintain all of them in one place.
> 
> Or if I am missing something, how are they different from the others?

i wanted some basic tests here so in case we change anything in future with respect to dex router , we shouldn't be bound to test it via cl-vault or running test from somewhere outside the scope so kept it very functional.

lubkoll:
> > Tests looks good, but I see redundancies with the ones we have on CL Vault contract. Maybe we should maintain all of them in one place.
> > Or if I am missing something, how are they different from the others?
> 
> i wanted some basic tests here so in case we change anything in future with respect to dex router , we shouldn't be bound to test it via cl-vault or running test from somewhere outside the scope so kept it very functional.

Imo, we should avoid test repetition where possible. Either we have it here or in the cl-vault. If it needs to be tested here, then the tests need to be here, else they need to be in the cl-vault repo.

smart-contracts/contracts/dex-router-osmosis/src/tests/initialize.rs
```diff
@@ -0,0 +1,483 @@
...
+#[cw_serde]
+pub struct PoolWithDenoms {
+    pub pool: u64,
+    pub denom0: String,
+    pub denom1: String,
+}
+
+pub fn default_init() -> (OsmosisTestApp, Addr, Vec<PoolWithDenoms>, SigningAccount) {
+    let app = OsmosisTestApp::new();
+    let tf = TokenFactory::new(&app);
```
magiodev, line 44: This can be probably be avoided, we can just init the account with 4 denoms in the balance so we save the creation, mint and assignation of such denoms we need for testing purposes.

smart-contracts/contracts/dex-router-osmosis/src/tests/initialize.rs
```diff
@@ -0,0 +1,483 @@
...
+                        ),
+                    },
+                ],
+                future_pool_governor: "overwritten".to_string(),
+            },
+        ],
+    )
+}
+
+pub fn init_test_contract(
```
magiodev, line 240: This fixture (as the others we have on cl-vault) looks very monolitic. Maybe we should start writing smaller helpers like create_cl_pool() as single one. create_gamm_pool() also as a single one. And another one to instantiate the contract which could probably be avoided as it would be a one-line invoke.

So in each test case we can create only the pool we need, and instantiate the contract accordingly to then do things such as SetPath, RemovePath and Swap. Swap will need liquidity so the admin should also create an initial position to facilitate the swaps in the happy cases we plan to write.

smart-contracts/contracts/dex-router-osmosis/src/tests/swap.rs
```diff
@@ -42,3 +46,95 @@ fn test_swap_with_too_many_funds_throws() {
     let err = execute(deps.as_mut(), env, info, msg).unwrap_err();
     assert_eq!(err, ContractError::Funds(FundsError::InvalidAssets(1)))
 }
+
+#[test]
+#[ignore]
+fn test_swap_over_singl_cl_route() {
```
magiodev, line 52: Typo here singl_ to single_

smart-contracts/contracts/dex-router-osmosis/src/tests/swap.rs
```diff
@@ -42,3 +46,95 @@ fn test_swap_with_too_many_funds_throws() {
...
+            },
+            &[Coin::new(10000u128, pools.first().unwrap().denom0.clone())],
+            &admin,
+        )
+        .unwrap();
+}
+
+#[test]
+#[ignore]
+fn test_swap_over_singl_gamm_route() {
```
magiodev, line 98: Same typo

smart-contracts/contracts/dex-router-osmosis/src/tests/initialize.rs
```diff
@@ -0,0 +1,383 @@
...
+
+    // Load compiled wasm bytecode
+    let wasm_byte_code = std::fs::read(filename).unwrap();
+    let code_id = wasm
+        .store_code(&wasm_byte_code, None, &admin)
+        .unwrap()
+        .data
+        .code_id;
+
+    // Instantiate vault
```
magiodev, line 329: Wrong copy pasted comment, this is the dex-router not the vault

magiodev (commented):
Looks mostly good, minor comments about typos. Also, I think it would be great to test a multihop swap across several pools.

We could run the fixture of creating the cl_poool or gamm_pool twice and swap using both.

Suggested cases are:
- Multihop cl / cl
- multihop gamm / gamm
- multihop cl / gamm
- multihop gamm/cl

This would be probably redundant as the type of pool is not concern of the swap router. If we agree with this last sentence probably we can also go for a single multihop test case.

magiodev (approved):

PR #639, opened 2024-06-28
Feat/multirange

LaurensKubat:
## 1. Overview

Adds support for multiple positions

TODO needs rest of description

## 2. Implementation details

The state of the contract is expended to support multiple positions. To manage multiple positions new strategist actions are added. These actions are:
- Create position; create a new position from all free funds in the contract balance
- Delete position; delete a position by withdrawing all liquidity
- Increase funds; Add free funds from the contract balance to a position
- Decrease funds; Remove liquidity from a position and return the funds to the contract balance
- MovePosition; Move a position from one range to another and swap any free funds and funds previous in the position to accommodate this new position

The contract itself uses a main position for users to deposit and withdraw against. 
- Deposit, A user deposits funds into the vault and gets vault shares in return. The user deposit is forced to be in the ratio of the main position. This allows the strategist to give a preference for a deposit in certain tokens. When the autocompounding endpoint is called, these funds are added to the main position when the autocompounding endpoint is called. This is similar to the single position version
- Withdraw, A user withdraws according to their vault shares. There are two ways for a user to currently withdraw their funds. The distinction between the two withdrawal methods depends on whether the main position has enough funds to accommodate the withdrawal. If there are enough funds in the main position, the user only receives funds from the main position. If this is not the case, user funds are removed pro rato from all positions and the free balance of the vault.

The first method exists to improve the efficiency of the multiple positions created by the strategist. If there are net 0 inflows and outflows of the vault, but still inflows and outflows happening, the vault should not drain the secondary positions created, but instead optimize this against the main position.
The second method ensure that a vault can be fully withdrawn permissionlessly without needing any interaction from the strategist.

## 4. Checklist


smart-contracts/contracts/cl-vault/src/state.rs
```diff
@@ -81,17 +81,24 @@ pub struct Position {
     pub claim_after: Option<u64>, // this should be off chain computed and set in order to avoid forfeiting incentives
 }
 
-pub const POSITION: Item<Position> = Item::new("position");
-
+// positions in the contract, the key should be the same as the position's id in Osmosis
+pub const POSITIONS: Map<u64, Position> = Map::new("positions");
+pub const MAIN_POSITION: Item<u64> = Item::new("main_position");
```
lubkoll, line 86:
```suggestion
pub const MAIN_POSITION_ID: Item<u64> = Item::new("main_position");
```

smart-contracts/contracts/cl-vault/src/state.rs
```diff
@@ -81,17 +81,24 @@ pub struct Position {
...
+pub const MERGE_MAIN_POSITION: Item<bool> = Item::new("merge_main_position");
+
+/// The queue of positions currently being merged
 pub const CURRENT_MERGE: Deque<CurrentMergeWithdraw> = Deque::new("current_merge");
 
 #[cw_serde]
 pub struct CurrentMergePosition {
     pub lower_tick: i64,
     pub upper_tick: i64,
+    pub claim_after: Option<u64>,
```
lubkoll, line 101: To make the unit visible in client code:
```suggestion
    pub claim_after_secs: Option<u64>,
```

smart-contracts/contracts/cl-vault/src/test_helpers.rs
```diff
@@ -241,6 +234,162 @@ pub fn mock_deps_with_querier_with_balance(
...
+pub fn mock_deps_with_querier_with_balance_with_positions(
+    info: &MessageInfo,
+    balances: &[(&str, &[Coin])],
+    main_position: FullPositionBreakdown,
+    secondary_positions: Vec<FullPositionBreakdown>,
+) -> OwnedDeps<MockStorage, MockApi, QuasarQuerier, Empty> {
+    let mut positions = secondary_positions.clone();
+    positions.extend_from_slice(&[main_position.clone()]);
+
+    let mut deps = OwnedDeps {
```
lubkoll, line 246: Would be nice to have a mock_deps function similar to the one in cosmwasm_std::testing to make that less verbose.

smart-contracts/contracts/cl-vault/src/test_tube/withdraw_main_pro_rato.rs
```diff
@@ -0,0 +1,401 @@
...
+                &QueryMsg::VaultExtension(ExtensionQueryMsg::ConcentratedLiquidity(
+                    ClQueryMsg::Positions {},
+                )),
+            )
+            .unwrap();
+        // since we are withdrawing from multiple different positions here in the second difference, we might encounter multiple
+        // moments where we round down, hence we can be off by 1 * the amount of positions + 1 for the free funds.
+        // since each of those calculations is a potential round down
+        let allowed_absolute_diff = positions.positions.len() as u128 + 1_u128;
+        assert_eq_with_diff!(
```
lubkoll, line 391: why not exact comparison?

smart-contracts/contracts/cl-vault/src/test_tube/asserts.rs
```diff
@@ -0,0 +1,52 @@
...
+macro_rules! assert_eq_with_diff {
+    ($left:expr, $left_tag:expr, $right:expr, $right_tag:expr, $max_rel_diff:expr, $max_absolute_diff:expr $(,)?) => {{
+        $crate::test_tube::asserts::assert_eq_with_diff_impl($left, $left_tag, $right, $right_tag, $max_rel_diff, $max_absolute_diff, None);
+    }};
+    ($left:expr, $left_tag:expr, $right:expr, $right_tag:expr, $max_rel_diff:expr, $max_absolute_diff:expr, $($args:tt)+) => {{
+        $crate::test_tube::asserts::assert_eq_with_diff_impl($left, $left_tag, $right, $right_tag, $max_rel_diff, $max_absolute_diff, Some(format!($($args)*)));
+    }};
+}
+
+/// assert that some two values are at most both the relative difference and absolute difference away from each other
```
lubkoll, line 15: Combining checks for relative and absolute difference is not a good idea. What is the underlying issue that is addressed here?

smart-contracts/contracts/cl-vault/src/vault/concentrated_liquidity.rs
```diff
@@ -48,35 +56,165 @@ pub fn create_position(
...
+    querier: &QuerierWrapper,
+    position_id: u64,
+) -> Result<FullPositionBreakdown, ContractError> {
     let cl_querier = ConcentratedliquidityQuerier::new(querier);
-    let position = cl_querier.position_by_id(position.position_id)?;
+    let position = cl_querier.position_by_id(position_id)?;
     position.position.ok_or(ContractError::PositionNotFound)
 }
 
+pub fn get_parsed_position(
```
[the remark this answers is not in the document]
LaurensKubat: A normal position is according to the Osmosis protobuf definitions, so that still contains a lot of `Option<>`, `String` instead of `Decimal`. All these `Option`s are not actually options from the Osmosis side but weird side effects of the osmosis-std builder. The Parsed position already takes care of that for us. This implementation parses an Osmosis struct

smart-contracts/packages/interface/src/range_middleware.rs
```diff
@@ -0,0 +1,28 @@
+use cw_orch::{interface, prelude::*};
+
+use range_middleware::msg::{ExecuteMsg, InstantiateMsg, MigrateMsg, QueryMsg};
+
+pub const CONTRACT_ID: &str = "range_middleware";
+
+#[interface(InstantiateMsg, ExecuteMsg, QueryMsg, MigrateMsg, id = CONTRACT_ID)]
+pub struct RangeMiddlewareContract;
+
+impl<Chain> Uploadable for RangeMiddlewareContract<Chain> {
```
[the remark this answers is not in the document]
LaurensKubat: Uploadable is the trait from cw-orch that defines how to upload a contract for different usecases. So when we use Osmosis test-tube, this trait defines where to find the built artifact

smart-contracts/contracts/cl-vault/src/test_tube/withdraw_main_pro_rato.rs
LaurensKubat: Everytime we calculate a percentage, we need to round down to make sure that the vault never "drains" itself (which rounding up will do). If we have 5 positions, we'd round down 5 times for the positions and 1 time for the free balance, because of which the user would get 5 less microtokens(uosmo etc). On the main position withdraw, the user would get 1 rounding down from the position. The tests needs to account for that difference

smart-contracts/contracts/cl-vault/src/test_tube/asserts.rs
LaurensKubat: This was previously used to ensure certain asserts at low token amounts where the absolute diff was bigger than the allowed relative diff still passed. For example if we assert with 100uosmo and 100uatom and we round down twice, we have a diff of 2%. So with this we can still test for approx 0.001% diffs on certain functions like swap functionality and also test low values

LaurensKubat: https://github.com/quasar-finance/quasar/pull/639#discussion_r1662596165 At this comment I do now think I should have used an comparison with only an allowed absolute diff, will change that

smart-contracts/contracts/cl-vault/src/test_tube/withdraw_main_pro_rato.rs
LaurensKubat: This can be rewritten to assert the `allowed_absolute_diff`, will change this

smart-contracts/contracts/cl-vault/src/test_helpers.rs
LaurensKubat: I don't think we can get rid of the verbosity here, because cosmwasm::testing doesn't export a nice constructor for `OwnedDeps`. We could improve the verbosity of test helpers a bit by changing `mock_deps_with_querier_with_balance_with_positions` into a builder that we can call similar. So the final call would be something like
`MockedDeps::new().with_querier().with_balance().with.positions().build()`. I'm not sure whether that's actually worth the effort though

smart-contracts/contracts/cl-vault/src/vault/concentrated_liquidity.rs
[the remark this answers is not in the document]
lubkoll: thx

smart-contracts/contracts/cl-vault/src/test_helpers.rs
lubkoll: I was just thinking about introducing a local function `mock_deps(positions, balances)` or options of the arguments. But rather a detail, so happy if you just close this if it is not relevant enough from your side

lubkoll:
- Any chance to split this up? This is just to big to review. When I open it in github most of the files are not even shown, so I have to click on each one individually to see it at all:(
- I am a bit confused that  this change requires to touch 79 files. Maybe we can clean up a bit in order to reduce that?
- I recommend to rebase against main asap, else the pain will just get bigger

LaurensKubat:
> Any chance to split this up? This is just to big to review. When I open it in github most of the files are not even shown, so I have to click on each one individually to see it at all:(

Let me see if I can move some changes to another PR. I don't have super high hopes though since there is quite a lot of refactoring that happened to support multiple positions

> I am a bit confused that this change requires to touch 79 files. Maybe we can clean up a bit in order to reduce that?

Let me take a look, I should be able to lower that. But to test the range middleware update I introduced cw-orch, so quite a lot of files were touched. I can move that to a seperate PR, but then we have a testless PR which is not what I'm looking for

>I recommend to rebase against main asap, else the pain will just get bigger

🫡

lubkoll (dismissed):

smart-contracts/Cargo.toml
```diff
@@ -15,6 +15,10 @@ members = [
   "contracts/cl-vault",
   "contracts/merkle-incentives",
   "contracts/range-middleware",
+  "contracts/cw-vault-standard",
+  "contracts/cw-vault-multi-standard",
+  "packages/interface"
```
lubkoll, line 20: I used the other packages as workspace-dependencies. Unfortunately, I don't recall the issue, but was having some in the past, so moved them out of this array. If there is not a strong argument for keeping it here, I'd prefer to also do that.

ajansari95:
Closing this in favour of  #650

PR #641, opened 2024-07-01
Ignore querier error in swap simulations

lubkoll:


smart-contracts/contracts/dex-router-osmosis/src/contract.rs
```diff
@@ -351,8 +357,13 @@ pub fn simulate_swaps(
     path: Vec<SwapAmountInRoute>,
 ) -> Result<Uint128, ContractError> {
     let querier = PoolmanagerQuerier::new(&deps.querier);
-    let response = querier.estimate_swap_exact_amount_in(0, offer.to_string(), path)?;
-    Ok(Uint128::from_str(&response.token_out_amount)?)
+    let response = querier.estimate_swap_exact_amount_in(0, offer.to_string(), path);
```
magiodev, line 360: Please leave a comment about the hardcoded pool_id 0 which could be no-sense for a non conteztualized reader.

smart-contracts/contracts/dex-router-osmosis/src/error.rs
```diff
@@ -31,6 +31,9 @@ pub enum ContractError {
 
     #[error("Can't set empty path.")]
     EmptyPath {},
+
+    #[error("Not enough liquidity for swap from {offer:?} to {ask_denom:?}")]
+    NotEnoughLiquidity { offer: Coin, ask_denom: String },
```
magiodev, line 36: Is that error only about liquidity issues? May it be for another reason so would make more sense a generic SwapError from asset to asset?

magiodev (commented):
Lgtm! Minor comments left

smart-contracts/contracts/dex-router-osmosis/src/contract.rs
lubkoll: done

smart-contracts/contracts/dex-router-osmosis/src/error.rs
lubkoll: done, gave the error a more generic name, other possible causes are
- osmosis messed up (though we shouldn't care about that in our implementation)
- a pool that is used in a path got removed

magiodev (approved):

PR #643, opened 2024-07-04
Add pre-commit hook for schema generation and formatting

lubkoll:


scripts/git/pre-commit
```diff
@@ -0,0 +1,52 @@
...
+
+  cargo fmt --all
+  cd ${REPO_ROOT}
+  git add $CHANGES
+fi
+
+CONTRACT_NAME=dex-router-osmosis
+if [[ $CHANGES == *contracts/${CONTRACT_NAME}* ]]
+then
+  echo "Changes in ${CONTRACT_NAME} contract"
```
magiodev, line 37: This looks kinda redundant respect ht eprevious one, also, is veruy specific. Can we somehow generalize the --feature flag and so iterate automatically all the `contracts/**` sub-folders and generate schema?

If iterating is too much, lets just hardcode contract names.

lubkoll: Introduced a function (now properly using `cargo schema` and `.cargo/config.toml`) to update schemas and fix formatting. 
For now I did hard-code the contracts that check. I also like the idea of automatically going over all contracts. Before I enable that, I would prefer to clean up the commented-out contracts, i.e. either remove them or enable them again.

magiodev (approved):

PR #649, opened 2024-07-12
feat: replace all admins with `mars-owner`

ajansari95:
## 1. Overview
In context of range-midddleware :
- admin , submitter admin and executor admins are migrated to use `mars-owner` instead

## 4. Checklist


smart-contracts/contracts/range-middleware/src/state.rs
```diff
@@ -12,3 +13,7 @@ pub struct NewRange {
 pub const PENDING_RANGES: Map<Addr, NewRange> = Map::new("pending_ranges");
 pub const RANGE_SUBMITTER_ADMIN: Item<Addr> = Item::new("range_submitter_admin");
```
LaurensKubat, line 14: Either we should keep the admin fields and migrate them, or Initialize new ones and clean these up so that our state stays clean

LaurensKubat: Since the migrate code already exists for this, we probably want to inline this or add a #[deprecate] so we properly deprecate these fields

LaurensKubat (changes requested):
small change needed to properly deprecate the old admin state items

smart-contracts/contracts/range-middleware/src/state.rs
lubkoll: imo, we should remove them completely, i.e.
- Remove the old items here
- in migrate read the values from locally defined `Item::new("...")` and clean the associated storage.

lubkoll (changes requested):
lgtm after fixing

smart-contracts/contracts/range-middleware/src/state.rs
ajansari95: resolved [69b1120](https://github.com/quasar-finance/quasar/pull/649/commits/69b1120dee0603a9f8f86a3c4c8208786b52dd8f)

smart-contracts/contracts/range-middleware/src/state.rs
```diff
@@ -10,5 +11,7 @@ pub struct NewRange {
 }
 
 pub const PENDING_RANGES: Map<Addr, NewRange> = Map::new("pending_ranges");
-pub const RANGE_SUBMITTER_ADMIN: Item<Addr> = Item::new("range_submitter_admin");
-pub const RANGE_EXECUTOR_ADMIN: Item<Addr> = Item::new("range_executor_admin");
+
+pub const OWNER: Owner = Owner::new("owner");
```
lubkoll, line 15: Why do we need a new owner here?

lubkoll (approved):

PR #657, opened 2024-07-17
CL Vault optimize range math operations

magiodev:
## 1. Overview

This PR aims to streamline and optimize the math operations related to `range.rs` in the scenario of `ModifyRange` message.

## 2. Implementation Details

Simplified the math operation on withdrawal of position reply in the context of re-ranging. Considering that on the reply, the balances are already incremented with the withdrawn assets from the previous step, it is unnecessary to subtract that with the withdrawn amount, only to re-add them with a `saturating_sub` in between.

Deprecated `CURRENT_BALANCE` state which was unnecessary. It was used before creating a position, and we were loading it to subtract the created position amounts from it. This has been replaced with a simple balances query which, in the context of a reply, already takes into account previous operations such as creating a position that already removes funds from the contract balance.

## 4. Checklist


magiodev:
Consider adding a `migrate()` entrypoint workflow to `.remove()` the `CURRENT_BALANCE` cosmwasm-storage item's state.

smart-contracts/contracts/cl-vault/src/vault/range.rs
```diff
@@ -189,6 +176,19 @@ pub fn handle_withdraw_position_reply(
         })
     }
 
+    Ok(AmountsAndTokensProvidedResponse {
```
lubkoll, line 179: Why do we return amount0 and amount1 if that is also provided through the tokens_provided?
Not sure about all the context, to me it looks like 
* either we should just return tokens_provided
* or we should just return amount0 and amount1 and construct tokens_provided elsewhere

smart-contracts/contracts/cl-vault/src/test_tube/helpers.rs
```diff
@@ -64,7 +64,7 @@ pub fn calculate_deposit_ratio(
         }
     }
 
-    // Calculate refunds
+    // Calculate refunds TODO check if this is correct
```
lubkoll, line 67: :D did you check already?

smart-contracts/contracts/cl-vault/src/vault/range.rs
magiodev: i.e. `do_swap_deposit_merge` function seeks for a tuple of both of them` (balances.amount0, balances.amount1)`. If we refactor all the other functions to work with tokens_provided Vec this would indeed be possible.

smart-contracts/contracts/cl-vault/src/helpers/getters.rs
```diff
@@ -312,183 +223,312 @@ pub fn get_unused_balances(querier: &QuerierWrapper, env: &Env) -> Result<CoinLi
...
-    additional_excluded_funds: (Uint128, Uint128),
-) -> Result<Decimal256, ContractError> {
-    // first get the ratio of token0:token1 in the position.
-    let p = get_position(deps.storage, &deps.querier)?;
-    // if there is no position, then we can assume that there are 0 unused funds
-    if p.position.is_none() {
-        return Ok(Decimal256::zero());
-    }
-    let position_unwrapped = p.position.ok_or(ContractError::MissingPosition {})?;
+// /// get_spot_price
```
lubkoll, line 259: and can we remove all the commented out code?

smart-contracts/contracts/cl-vault/src/vault/range.rs
magiodev: Addressed by splitting the getter helper in two different functions. So i got rid of the new struct i was introducing. Looks cleaner now, thanks!

lubkoll (dismissed):

PR #666, opened 2024-07-18
[CI] Add reusable workflow for test-tube tests

lubkoll:


smart-contracts/contracts/dex-router-osmosis/.cargo/config.toml
```diff
@@ -2,5 +2,5 @@
 wasm = "build --release --target wasm32-unknown-unknown"
 unit-test = "test --lib"
 schema = "run --bin schema"
-test-tube = "test --package dex-router-osmosis --lib -- --include-ignored test_tube:: --nocapture --test-threads=1"
```
magiodev, line 5: `--test-threads=1`  is needed due to some race condition that was causing not deterministic test outcomes.

smart-contracts/justfile
```diff
@@ -34,4 +34,11 @@ workspace-optimize:
             --mount type=volume,source="$(basename "$(pwd)")_cache",target=/target \
             --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
             --platform linux/amd64 \
-            cosmwasm/workspace-optimizer:0.16.0; fi
+            cosmwasm/optimizer:0.16.0; fi
+
+optimize-single-contract:
```
magiodev: I would call that `contract-optimize` to maintain some consistency with the above command.

magiodev (changes requested):

smart-contracts/contracts/dex-router-osmosis/.cargo/config.toml
lubkoll: added back

smart-contracts/justfile
lubkoll: removed, that did slip through from previous tests.
While that is convenient, we don't need it here

magiodev (approved):

PR #673, opened 2024-07-18
[CI] Run proptests only on main

lubkoll:


LaurensKubat (dismissed):
lgtm

smart-contracts/contracts/cl-vault/.cargo/config
```diff
@@ -2,5 +2,6 @@
 wasm = "build --release --lib --target wasm32-unknown-unknown"
 unit-test = "test --lib"
 schema = "run --bin schema"
-test-tube = "test --package cl-vault --lib -- --include-ignored test_tube:: --nocapture --test-threads=1"
+test-tube = "test --package cl-vault --lib -- --include-ignored test_tube:: --nocapture --test-threads=1 --skip test_complete_works"
```
LaurensKubat, line 5: This doesn't run the proptests. The `--skip` here skips the prop test, but the cl_vault.ym uses the rust-test-tube workflow which calls this command

LaurensKubat (changes requested):
The test-tube alias does not run the prop tests but is called by the workflow that should run the prop test

smart-contracts/contracts/cl-vault/.cargo/config
lubkoll: Yes, exactly that is the feature:)
I.e. I'd like to not run the proptest when we call `cargo test-tube`, but only on `cargo prop-test`. The problem is that currently this one test takes most of the time in the pre-commit verification, but brings close to 0 value (didn't see any failure of it triggering any action since I am here -- and it fails regularly with the code even remotely touching anything related to it). Even if it fails it doesn't give you much directly actionable input. 
My actual thoughts are that:
* we should kill the proptests
* else run it in a nightly job and just check if it is failing in the morning
* first step is though to get it out of pre-commit verification

LaurensKubat: That makes sense, in that case I was confused by the PR title, since this PR effectively removes prop tests from the CI. Also lets discuss proptests later over slack or smth

LaurensKubat (approved):
lgtm, please make sure to change the PR title or the commit, so we clearly see this PR removes proptests from the CI

smart-contracts/contracts/cl-vault/.cargo/config
lubkoll: thanks for catching. That shouldn't be the case and running the proptest on main was part of this before I had to rebase 100x as there were no reviews.

lubkoll: added it back

ajansari95 (approved):
lgtm

PR #674, opened 2024-07-19
Feat/cl vault range calc swap

magiodev:
## 1. Overview

This PR optimizes the way we create the next position during a ModifyRange operation, removing dangerous math operations by reading contract balances instead, leveraging the fact that in the reply we already updated balances after the reply_on_success has been called.

## 4. Checklist

### DIFF
--- a/smart-contracts/contracts/cl-vault/src/state.rs
+++ b/smart-contracts/contracts/cl-vault/src/state.rs
@@ -24,6 +24,9 @@ pub const USER_REWARDS: Map<Addr, CoinList> = Map::new("user_rewards");
 #[deprecated]
 pub const CURRENT_BALANCE: Item<(Uint128, Uint128)> = Item::new("current_balance"); // CURRENT_BALANCE is intended as CURRENT_SWAP_BALANCE
 
+#[deprecated]
+pub const CURRENT_SWAP: Item<(SwapDirection, Uint128)> = Item::new("current_swap");
+
 /// metadata useful for display purposes
 #[cw_serde]
 pub struct Metadata {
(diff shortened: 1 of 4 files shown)


smart-contracts/contracts/cl-vault/src/state.rs
```diff
@@ -24,6 +24,9 @@ pub const USER_REWARDS: Map<Addr, CoinList> = Map::new("user_rewards");
 #[deprecated]
 pub const CURRENT_BALANCE: Item<(Uint128, Uint128)> = Item::new("current_balance"); // CURRENT_BALANCE is intended as CURRENT_SWAP_BALANCE
 
+#[deprecated]
+pub const CURRENT_SWAP: Item<(SwapDirection, Uint128)> = Item::new("current_swap");
```
lubkoll, line 28: what about removing this completely and adding
`Item::new("current_swap").remove(...` in the migrate msg?

magiodev: Yeah we thought that but then there would be a lot of refactoring on the migrate tests side of things and we prefer to leave things as this, avoid unnecessary changes and clean this out completely in the next iteration. What do you think?

ajansari95 (approved):
lgtm

smart-contracts/contracts/cl-vault/src/state.rs
lubkoll: There shouldn't be a lot of refactoring. I don't talk about changing code that was changed before, but only about the one variable of this PR, the CURRENT_SWAP. So the required changes would only be:
* remove it here
* in migrate(, replace `CURRENT_SWAP` with `Item::new("current_swap")`
* add the import for Item

The motivation is that we already pile up deprecated variables (and hide all deprecation warnings). I don't see any advantage that we get from doing that, it just seems to add noise. The mentioned approach avoids that and makes it easier to cleanup after migration (i.e. only the content of migrate needs to be cleared).

lubkoll (approved):

PR #687, opened 2024-07-22
Add boilerplate for babylon vault

lubkoll:


smart-contracts/Cargo.toml
```diff
@@ -17,6 +17,7 @@ members = [
   "contracts/dex-router-osmosis",
   "contracts/lst-dex-adapter-osmosis",
   "contracts/lst-adapter-osmosis",
+  "contracts/babylon-vault",
 ]
 
 [workspace.dependencies]
```
ajansari95, line 23: I dont think it makes sense to use the same workspace as osmosis-contracts as the target chain isn't osmosis

ajansari95 (commented):
I think we can start segregating workspaces based on Product. In here the workspace shouldn't be shared with osmosis contracts

smart-contracts/Cargo.toml
lubkoll: We had discussed that when starting with the monorepo (one workspace, different suffixes). If there is not a very strong reason to do so, I would prefer to not open that bottle again.

ajansari95: this worksapce has osmosis dependencies, i think we decided product based worksapces cause target chains are different

smart-contracts/contracts/babylon-vault/src/msg.rs
```diff
@@ -0,0 +1,32 @@
+use cosmwasm_schema::{cw_serde, QueryResponses};
+use cosmwasm_std::{Timestamp, Uint128};
+
+#[cw_serde]
+pub struct InstantiateMsg {}
+
+#[cw_serde]
+pub enum ExecuteMsg {
```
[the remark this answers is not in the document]
lubkoll: did look into it. Not convinced yet. I see a point in the unification of interfaces. As a negative, that leads to unimplemented public messages, i.e. I didn't see an option to remove the Redeem message.
Currently I prefer to keep things as they are, as it is easier to add the cw_vault_standard later,
* either through an adapter contract
* or by replacing the interface of the babylon vault.

smart-contracts/Cargo.toml
arhamchordia: Makes sense to change it later once we have more segregations in chains. Right now everything in same workspace is okay.

But in general if we are prepping to move towards mono-repo (multiple destination contracts) discussion we had then shouldn't we start it now? Might be a first step in right direction.

lubkoll: Will create two separate workspaces osmosis and babylon and then continue on this PR.

lubkoll: done

lubkoll:
Closing the PR for now, keeping the branches. We can reopen that when we actually do it

PR #690, opened 2024-07-22
Fix/makefile

arhamchordia:
## 1. Overview

Restructuring makefile to keep make commands more readable and easy to execute.

## 2. Implementation details

Adding different makefiles in scripts/makefiles
Separating them into : 
1. build.mk
2. docker.mk 
3. and so on....

## 3. How to test/use

Run commands one by one to test them.


arhamchordia:
Seems like my conflicts were not resolved

PR #692, opened 2024-07-22
Refactoring makefile.

arhamchordia:
## 1. Overview

Restructuring makefile to keep make commands more readable and easy to execute.

## 2. Implementation details

Adding different makefiles in scripts/makefiles
Separating them into : 
1. build.mk
2. docker.mk 
3. and so on....

## 3. How to test/use

Run commands one by one to test them.

### DIFF
--- a/scripts/makefiles/proto.mk
+++ b/scripts/makefiles/proto.mk
@@ -1,6 +1,10 @@
 ###############################################################################
 ###                         Proto & Mock Generation                         ###
 ###############################################################################
+protoVer=0.13.0
+protoImageName=ghcr.io/cosmos/proto-builder:$(protoVer)
+protoImage=$(DOCKER) run --rm -v $(CURDIR):/workspace --workdir /workspace $(protoImageName)
+
 proto-help:
 	@echo "proto subcommands"
 	@echo ""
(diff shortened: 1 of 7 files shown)


scripts/makefiles/proto.mk
```diff
@@ -0,0 +1,35 @@
...
+	@echo "Available Commands:"
+	@echo "  all        Run proto-format and proto-gen"
+	@echo "  gen        Generate Protobuf files"
+	@echo "  gen-1      Generate Protobuf files (old relic)"
+	@echo "  doc        Generate proto docs"
+
+proto: proto-help
+proto-all: proto-gen
+
+# todo : @AJ needs to address this after removing third_party. Refer this for removal https://github.com/osmosis-labs/osmosis/blob/188abfcd15544ca07d468c0dc0169876ffde6079/scripts/makefiles/proto.mk#L39
```
magiodev, line 19: Is this TODO for now?

magiodev (commented):
Refactor looks good. I see lot of redundancy over commands, should we try to reduce the amount of repeated or similar ones?

akure (commented):
The structure is good. Is there any way to test all the Makefile commands in one go in local.  Like a command that will build, test and try maximum possibilities.  Not necessarily to be part of this PR.

scripts/makefiles/proto.mk
arhamchordia: Not for now. To be addressed by @ajansari95 in his cleanup PR

scripts/makefiles/proto.mk
```diff
@@ -0,0 +1,35 @@
...
+proto-gen:
+	@echo "Generating Protobuf files"
+	@sh ./scripts/protocgen.sh
+
+# todo : @AK need the reason why it was there earlier
+proto-gen-1:
```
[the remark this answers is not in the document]
arhamchordia: Sure. So we can get rid of it now?

arhamchordia:
> The structure is good. Is there any way to test all the Makefile commands in one go in local. Like a command that will build, test and try maximum possibilities. Not necessarily to be part of this PR.

Yes. We can make a general command in makefile that includes all of them together. Something like `make execute-all` which in turn executes everything.

arhamchordia:
> Refactor looks good. I see lot of redundancy over commands, should we try to reduce the amount of repeated or similar ones?

I see only distroless build as the redundant one. Any others that you could find?

akure (approved):
lg

PR #701, opened 2024-07-23
switch `swap-non-vault-funds` entrypoint  permissionless

ajansari95:
## 1. Overview
- make the `swap_non_vault_funds` permissionless.
- idea behind this is so that the swap process can be automated and thus efficiently autocompounded (via cronjob)

## 2. Implementation details
- remove the assertion from range admin

## 4. Checklist
- NA

### DIFF
--- a/smart-contracts/osmosis/contracts/cl-vault/src/contract.rs
+++ b/smart-contracts/osmosis/contracts/cl-vault/src/contract.rs
@@ -137,7 +137,7 @@ pub fn execute(
                     )?,
                 ),
                 crate::msg::ExtensionExecuteMsg::SwapNonVaultFunds { swap_operations } => {
-                    execute_swap_non_vault_funds(deps, env, info, swap_operations)
+                    execute_swap_non_vault_funds(deps, env.contract.address, swap_operations)
                 }
                 crate::msg::ExtensionExecuteMsg::CollectRewards {} => {
                     execute_collect_rewards(deps, env)
--- a/smart-contracts/osmosis/contracts/cl-vault/src/helpers/msgs.rs
+++ b/smart-contracts/osmosis/contracts/cl-vault/src/helpers/msgs.rs
@@ -58,7 +58,11 @@ pub fn refund_bank_msg(
 
 /// swap will always swap over the CL pool. In the future we may expand the
 /// feature such that it chooses best swaps over all routes
-pub fn swap_msg(deps: &DepsMut, env: &Env, params: SwapParams) -> Result<CosmosMsg, ContractError> {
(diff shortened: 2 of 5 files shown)


smart-contracts/contracts/cl-vault/src/vault/swap.rs
```diff
@@ -85,6 +76,7 @@ pub fn execute_swap_non_vault_funds(
             vault_config.swap_max_slippage.denominator(),
         )?;
 
+        // Add swap messages
```
lubkoll, line 79: this comment does not provide more information than the following code, plz remove

smart-contracts/contracts/cl-vault/src/vault/swap.rs
```diff
@@ -29,53 +28,45 @@ pub struct SwapParams {
...
-            return Err(ContractError::InsufficientFunds {});
+        if balance_in_contract.is_zero() {
+            return Err(ContractError::InsufficientFundsForSwap {
+                balance: balance_in_contract,
+                needed: Uint128::new(1),
+            });
         }
 
-        // TODO_FUTURE: We could be swapping into the actual vault balance so we could prepend_swap() the autocompound entrypoint.
+        // Split balance into two parts
```
lubkoll, line 65: this comment does not provide more information than the following code, plz remove

smart-contracts/contracts/cl-vault/src/vault/swap.rs
```diff
@@ -85,6 +76,7 @@ pub fn execute_swap_non_vault_funds(
             vault_config.swap_max_slippage.denominator(),
         )?;
 
+        // Add swap messages
         swap_msgs.push(swap_msg(
```
lubkoll, line 80: I understand that this is not part of your change, so no need to address this now. If you feel motivated to (and it doesn't lead to many other files being changed), then it would be nice to not use `env` in the `swap_msg` function, but just provide the contract address (the only part of env that is really used)

smart-contracts/contracts/cl-vault/src/vault/swap.rs
```diff
@@ -29,53 +28,45 @@ pub struct SwapParams {
 pub fn execute_swap_non_vault_funds(
     deps: DepsMut,
     env: Env,
-    info: MessageInfo,
+    _info: MessageInfo,
```
lubkoll, line 31: plz remove unused var

lubkoll (changes requested):

lubkoll (approved):

magiodev (changes requested):
This is wrong

PR #702, opened 2024-07-23
fix: refactor deposit logic

ajansari95:
## 1. Overview
- Simplified deposit logic by consolidating common code and eliminating redundancy  wherever possible.

## 2. Implementation details
- Optimized AnyDeposit logic by resolving repetitive if-else blocks.
- Unified deposit types using an enum, removing separate deposit modules

## 4. Checklist


smart-contracts/contracts/cl-vault/src/vault/deposit.rs
```diff
@@ -84,24 +318,84 @@ pub(crate) fn execute_exact_deposit(
...
+    pool_config: PoolConfig,
+    swap_direction: SwapDirection,
+    token_in_amount: Uint128,
+    swappable_amount: (Uint128, Uint128),
+    deposit_amount_in_ratio: (Uint128, Uint128),
+    max_slippage: Decimal,
+    recipient: &Addr,
+) -> Result<SwapCalculationResult, ContractError> {
+    // TODO check that this math is right with spot price (numerators, denominators) if taken by legacy gamm module instead of poolmanager
+    // TODO check on the twap_window_seconds (taking hardcoded value for now)
```
arhamchordia, line 337: Is this todo for this PR?

smart-contracts/contracts/cl-vault/src/vault/deposit.rs
```diff
@@ -303,14 +597,15 @@ mod tests {
         //     )
         //     .unwrap();
```
arhamchordia, line 598: Remove commented code.

magiodev:
I don't quite understand why the rename from `any_deposit` to `deposit.rs`. This seems unnecessary and much less consistent. If we want a single `deposit.rs` file, we should also merge in the logic from `exact_deposit.rs`, but I tend to prefer keeping things separate. This is just your preference, but in my opinion, it should be one or the other.

magiodev (commented):
I don't quite understand why the rename from `any_deposit` to `deposit.rs`. This seems unnecessary and much less consistent. If we want a single `deposit.rs` file, we should also merge in the logic from `exact_deposit.rs`, but I tend to prefer keeping things separate. This is just your preference, but in my opinion, it should be one or the other.

magiodev (changes requested):
After checking out the branch locally I realized that you did hehe, files changes on Github is not showing any change on exact_deposit.rs file. Anyway, I still think that merging those two in one becomes monolitic and unnecessary.

smart-contracts/contracts/cl-vault/src/vault/deposit.rs
ajansari95: it was added earlier during debugging accounting error, removed.

smart-contracts/osmosis/contracts/cl-vault/src/msg.rs
```diff
@@ -173,6 +173,11 @@ pub struct InstantiateMsg {
     pub initial_upper_tick: i64,
 }
 
+pub enum DepositType {
```
LaurensKubat, line 176: Is this distinction not already made by the two different entrypoints?

smart-contracts/osmosis/contracts/cl-vault/src/vault/deposit.rs
```diff
@@ -1,42 +1,74 @@
...
-    // get the amount of funds we can deposit from this ratio
     let (deposit, refund): ((Uint128, Uint128), (Uint128, Uint128)) =
         get_depositable_tokens(deps.branch(), token0.clone(), token1.clone())?;
 
-    let vault_denom = VAULT_DENOM.load(deps.storage)?;
-    let total_vault_shares: Uint256 = query_total_vault_token_supply(deps.as_ref())?.total.into();
-
-    let user_value = get_asset0_value(deps.storage, &deps.querier, deposit.0, deposit.1)?;
-    let refund_value = get_asset0_value(deps.storage, &deps.querier, refund.0, refund.1)?;
+    // NOTE: both the deposit arms here have some repeated math logic,
```
LaurensKubat, line 46: If we combine both entrypoints into a single function call, why do we still have math logic that is repeated?

LaurensKubat (changes requested):
The addition of the Deposit type enum does not make a whole lot of sense to me since that same information is already implied by the two different entry points. For this PR, I'd remove any changes related to that enum's addition and discuss that in slack and restrict the PR to the math refactor in range.rs

PR #710, opened 2024-07-25
Add vault share token

lubkoll:


smart-contracts/quasar/contracts/babylon-vault/src/contract.rs
```diff
@@ -26,7 +32,24 @@ pub fn instantiate(
...
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn reply(deps: DepsMut, _env: Env, reply: Reply) -> VaultResult {
+    match reply.id {
+        CREATE_DENOM_REPLY_ID => {
+            let response: MsgCreateDenomResponse = reply.result.try_into()?;
+            VAULT_DENOM.save(deps.storage, &response.new_token_denom)?;
+
+            Ok(Response::new().add_attribute("vault_denom", response.new_token_denom))
```
arhamchordia, line 49: I was wondering if we can standardise this structure of adding event and then attached attribute to it's events. So that it becomes easier to track later on while we index our contracts. 

Like this can be : `Ok(Response::new().add_event(Event::new("create-vault-denom").add_attributes(attributes)))`

Faced an issue earlier while indexing our contracts on Osmosis. @magiodev cleared it.

smart-contracts/quasar/contracts/babylon-vault/src/contract.rs
```diff
@@ -26,7 +32,24 @@ pub fn instantiate(
         mars_owner::OwnerInit::SetInitialOwner { owner: msg.owner },
     )?;
     set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
-    Ok(Response::default())
+    let msg = MsgCreateDenom {
+        sender: env.contract.address.to_string(),
+        subdenom: msg.subdenom,
+    };
+    Ok(Response::new().add_submessage(SubMsg::reply_on_success(msg, CREATE_DENOM_REPLY_ID)))
```
arhamchordia, line 39: Here we should add instantiate event and it's attributes.

arhamchordia (changes requested):

smart-contracts/quasar/contracts/babylon-vault/src/contract.rs
lubkoll: will open a discussion in slack later or tomorrow. Sounds to me like we need to agree on some convention for that

LaurensKubat: For creating/instantiating a vault, we should discuss whether we automatically start indexing that or need to manually add that. manually adding sounds most attractive imo since we don't want to or need to always use indexing resources right now for our current product.
Indexing of the actual token and it's holders should probably be done through the bank module. The current CL vaults keeps the shares and track user shares internally, If we hand out the users shares to the actual users instead of the cl vault, we can simply index token holders of the tokenfactory token

arhamchordia (approved):
From the above discussions this looks good to me.

PR #712, opened 2024-07-25
Feat/pre v50 cleanup

arhamchordia:
## 1. Overview

This PR aims to cleanup code for v50 upgrade.

## 2. Implementation details

Cleanup of code contains the following changes : 
1. Complete removal of qoracle, qtransfer and qvesting modules.
2. Removal of 3rd party proto directory.
3. Testutil cleanup.
4. Wasmbinding cleanup.
5. Lint checks.
6. Renaming quasarnoded to quasard.

## 3. How to test/use

Lint test can be done by running `make lint-all` 
Tests can be run by `make test-all`

## 4. Future Work (optional)

@ajansari95 needs to do the api changes in the later PRs


akure (changes requested):
As we cleaning, we should also clean demo directory , we can remove all lp , qoracle, band chain related files.

akure:
Also, as we editing the name of the binary the default home directory will also change, which will cause path issue during chain upgrade if any node is using the default path, this also need to be taken care. And Clear readme instructions should be added.

arhamchordia:
Yes yes. Makes total sense. I think we can let the name of the directory be the same old one. Just name change would be okay?

akure (approved):
lgtm

PR #714, opened 2024-07-25
WIP: Add deposits and withdrawals

lubkoll:
Still contains a dummy placeholder for slinky. Will update as soon as I have the messages for that.


lubkoll:
Closing the PR for now, keeping the branches. We can reopen that when we actually do it

PR #715, opened 2024-07-26
Swap non vault funds

magiodev:
## 1. Overview

As the PR has grown a bit too much, here is a list of the file changes that have been logically impacted, and the ones that have been simply renamed and do not contain any logical changes.

No logical file changes:
- smart-contracts/osmosis/contracts/cl-vault/src/contract.rs
- smart-contracts/osmosis/contracts/cl-vault/src/error.rs
- smart-contracts/osmosis/contracts/cl-vault/src/lib.rs
- smart-contracts/osmosis/contracts/cl-vault/src/instantiate.rs
- smart-contracts/osmosis/contracts/cl-vault/src/msg.rs
- smart-contracts/osmosis/contracts/cl-vault/src/test_helpers.rs
- smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/any_deposit.rs
- smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/setup.rs
- smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/authz.rs

Logical file changes:
- smart-contracts/osmosis/contracts/cl-vault/src/helpers/getters.rs
- smart-contracts/osmosis/contracts/cl-vault/src/helpers/msgs.rs
- smart-contracts/osmosis/contracts/cl-vault/src/vault/deposit.rs
- smart-contracts/osmosis/contracts/cl-vault/src/vault/range.rs
- smart-contracts/osmosis/contracts/cl-vault/src/vault/swap.rs
- smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/autocompound.rs

## 4. Checklist


smart-contracts/osmosis/contracts/cl-vault/src/vault/swap.rs
```diff
@@ -60,29 +61,31 @@ pub fn execute_swap_non_vault_funds(
...
-        // Calculate the minimum amount of tokens to be received
-        let token_out_min_amount_0 = part_0_amount.checked_multiply_ratio(
-            vault_config.swap_max_slippage.numerator(),
-            vault_config.swap_max_slippage.denominator(),
-        )?;
-        let token_out_min_amount_1 = part_1_amount.checked_multiply_ratio(
-            vault_config.swap_max_slippage.numerator(),
-            vault_config.swap_max_slippage.denominator(),
-        )?;
+        // TODO: Make this dynamic based on the current position's balancing
```
magiodev, line 64: This TODO

smart-contracts/osmosis/contracts/cl-vault/src/vault/swap.rs
```diff
@@ -60,29 +61,31 @@ pub fn execute_swap_non_vault_funds(
...
+        let half_balance_amount = balance_in_contract.checked_div(Uint128::new(2))?;
+
+        // Get twap price
+        // TODO: Think how we want to pass the twap_window_seconds that now is hardcoded to 0
```
magiodev, line 68: This TODO

smart-contracts/osmosis/contracts/cl-vault/src/vault/swap.rs
```diff
@@ -60,29 +61,31 @@ pub fn execute_swap_non_vault_funds(
...
+        let twap_price = get_twap_price(deps.storage, &deps.querier, &env, 0)?;
+        let token_out_min_amount_0 = half_balance_amount
+            .checked_multiply_ratio(twap_price.numerator(), twap_price.denominator())? // twap
+            .checked_multiply_ratio(
+                vault_config.swap_max_slippage.numerator(),
+                vault_config.swap_max_slippage.denominator(),
+            )?; // slippage
+        let token_out_min_amount_1 = half_balance_amount
+            .checked_multiply_ratio(twap_price.denominator(), twap_price.numerator())? // twap
+            .checked_multiply_ratio(
```
magiodev, line 78: Consider the config as 0.005. Check this math that looks wrong.

smart-contracts/osmosis/contracts/cl-vault/src/vault/swap.rs
```diff
@@ -91,10 +94,10 @@ pub fn execute_swap_non_vault_funds(
         )?);
```
magiodev, line 94: If we want this entrypoint to be permissionless, here we should probably hardcode the `forced_swap_route` to None, in order to leave the Dex Router work with its beth path.

magiodev: This is already covered by the swap_max_slippage field checking the token_out_min_amount. So we don't care if someone is hijacking the route passes through esoteric pool ids.

smart-contracts/osmosis/contracts/cl-vault/src/vault/range.rs
```diff
@@ -93,28 +91,19 @@ pub fn execute_update_range_ticks(
 ) -> Result<Response, ContractError> {
     assert_range_admin(deps.storage, &info.sender)?;
 
-    // todo: prevent re-entrancy by checking if we have anything in MODIFY_RANGE_STATE (redundant check but whatever)
+    // prevent re-entrancy by checking if we have anything in MODIFY_RANGE_STATE
+    if MODIFY_RANGE_STATE.may_load(deps.storage)?.is_some() {
```
lubkoll, line 95: how can we ever get into that situation?

smart-contracts/osmosis/contracts/cl-vault/src/vault/deposit.rs
```diff
@@ -203,27 +187,25 @@ fn execute_deposit(
     deps: &mut DepsMut,
     env: Env,
     recipient: Addr,
-    deposit: (Uint128, Uint128),
-    refund: (Coin, Coin),
+    deposit: (Uint128, Uint128), // TODO: This could be DepositInfo struct
```
magiodev, line 190: TODO

smart-contracts/osmosis/contracts/cl-vault/src/vault/swap.rs
```diff
@@ -1,120 +1,139 @@
...
-            vault_config.swap_max_slippage.denominator(),
+        // Get the current position balance ratio to compute the amount of external funds we want to swap into either token0 or token1 from the vault's pool
+        // TODO: This new get_position_balance helper looks redundant with get_depositable_tokens_impl, can we reuse this instead?
+        let position_balance = get_position_balance(deps.storage, &deps.querier)?;
+        let to_token0_amount =
+            Uint128::from((balance_in_contract.u128() as f64 * position_balance.0) as u128); // balance * ratio computed by current position balancing
+        let to_token1_amount =
+            Uint128::from((balance_in_contract.u128() as f64 * position_balance.1) as u128); // balance * ratio computed by current position balancing
+
+        // TODO: Validate that the swap_operation.pool_id_0 is about token_in_denom and pool_config.token0 assets or throw error
```
magiodev, line 88: TODO

smart-contracts/osmosis/contracts/cl-vault/src/vault/swap.rs
```diff
@@ -1,120 +1,139 @@
...
+                    amount: to_token0_amount,
+                },
+                vault_config.swap_max_slippage,
+                swap_operation.forced_swap_route_token_0,
+                twap_price_token_0,
+            )?
+            .swap_msg,
+        );
+
+        // TODO: Validate that the swap_operation.pool_id_1 is about token_in_denom and pool_config.token1 assets or throw error
```
magiodev, line 113: TODO

smart-contracts/osmosis/contracts/cl-vault/src/vault/swap.rs
```diff
@@ -1,120 +1,139 @@
...
-        let part_1_amount = balance_in_contract
-            .checked_add(Uint128::new(1))?
-            .checked_div(Uint128::new(2))?;
-
-        // TODO_FUTURE: We should be passing the max_slippage from outside as we do during ModifyRange
-        let token_out_min_amount_0 = part_0_amount.checked_multiply_ratio(
-            vault_config.swap_max_slippage.numerator(),
-            vault_config.swap_max_slippage.denominator(),
+        // Get the current position balance ratio to compute the amount of external funds we want to swap into either token0 or token1 from the vault's pool
+        // TODO: This new get_position_balance helper looks redundant with get_depositable_tokens_impl, can we reuse this instead?
```
magiodev, line 81: TODO

smart-contracts/osmosis/contracts/cl-vault/src/vault/swap.rs
```diff
@@ -123,72 +142,121 @@ pub fn execute_swap_non_vault_funds(
...
-            &pool_config.token0,
+    let pool_config = POOL_CONFIG.load(deps.storage)?;
+
+    // Determine the target token and amount based on swap direction
+    let (denom_out, amount_out_ratio) = match swap_direction {
+        SwapDirection::ZeroToOne | SwapDirection::AnyToOne => (
             &pool_config.token1,
-            token_in_amount
-                .checked_multiply_ratio(twap_price.numerator(), twap_price.denominator()),
+            (twap_price.numerator(), twap_price.denominator()), // TODO: Check if this is correct order for AnyToOne
```
magiodev, line 211: TODO

smart-contracts/osmosis/contracts/cl-vault/src/vault/swap.rs
```diff
@@ -123,72 +142,121 @@ pub fn execute_swap_non_vault_funds(
...
         ),
-        SwapDirection::OneToZero => (
-            &pool_config.token1,
+        SwapDirection::OneToZero | SwapDirection::AnyToZero => (
             &pool_config.token0,
-            token_in_amount
-                .checked_multiply_ratio(twap_price.denominator(), twap_price.numerator()),
+            (twap_price.denominator(), twap_price.numerator()), // TODO: Check if this is correct order for AnyToZero
```
magiodev, line 215: TODO

smart-contracts/osmosis/contracts/cl-vault/src/vault/range.rs
magiodev: we cant, im about to deprecate this. but this was a TODO there since a while. I guess the initial intention was something like the openzeppeling reentrancyGuard https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/ReentrancyGuard.sol

lubkoll: I see. Imo, reentrancy guards in cosmwasm contracts are a sign that one does not really know what's going on. There is no re-entrancy risk and caches that are consumed in replies are under our full control, so should always be empty after a tx completed.

smart-contracts/osmosis/contracts/cl-vault/src/vault/deposit.rs
```diff
@@ -118,33 +93,41 @@ pub(crate) fn execute_any_deposit(
...
+        &deps,
         &env,
-        pool_config,
         swap_direction,
-        swap_amount,
+        token_in.clone(),
         max_slippage,
-        None, // TODO: check this None
-        24u64,
+        None,
```
magiodev, line 114: Should we pass a default one?

smart-contracts/osmosis/contracts/cl-vault/src/helpers/getters.rs
```diff
@@ -38,26 +37,56 @@ pub fn get_asset0_value(
...
     storage: &dyn Storage,
     querier: &QuerierWrapper,
-    env: &Env,
+) -> Result<(f64, f64), ContractError> {
+    let position = get_position(storage, querier)?;
+    let asset0_amount = Uint128::from_str(&position.clone().asset0.unwrap_or_default().amount)?;
+    let asset1_amount = Uint128::from_str(&position.clone().asset1.unwrap_or_default().amount)?;
+
+    // Handle cases where either asset amount is zero
+    if asset0_amount.is_zero() && asset1_amount.is_zero() {
```
magiodev, line 58: This is probably unnecessary as we are not making checked operations

smart-contracts/osmosis/contracts/cl-vault/src/helpers/getters.rs
```diff
@@ -27,8 +27,7 @@ pub fn get_range_admin(deps: Deps) -> Result<Addr, ContractError> {
 pub fn get_asset0_value(
     storage: &dyn Storage,
     querier: &QuerierWrapper,
-    token0: Uint128,
-    token1: Uint128,
+    tokens_provided: (Uint128, Uint128),
```
lubkoll, line 30: I would prefer to keep the arguments as before and rename the function to `get_token0_value`. Providing pairs is not very readable and in this case relies on the the implicit convention that asset0 refers to the first entry in tokens_provided.

smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/autocompound.rs
```diff
@@ -287,8 +288,8 @@ fn test_autocompound_with_rewards_swap_non_vault_funds() {
         &ExecuteMsg::VaultExtension(ExtensionExecuteMsg::SwapNonVaultFunds {
             swap_operations: vec![SwapOperation {
                 token_in_denom: DENOM_REWARD.to_string(),
-                pool_id_0: swap_pools_ids[2],
-                pool_id_1: swap_pools_ids[1],
+                pool_id_token_0: swap_pools_ids[2],
```
lubkoll, line 291: I don't think that adding token to the name does provide additional information. So would prefer to either keep the old naming or replace the _0, _1-prefixes. If that is referring to pool tokens, then I'd prefer to use base and quote as pre- or suffix as these are most widely used (though I always confuse them so we would need to document that:) and have a clear meaning

smart-contracts/osmosis/contracts/cl-vault/src/vault/deposit.rs
lubkoll: imo, no

smart-contracts/osmosis/contracts/cl-vault/src/helpers/getters.rs
lubkoll: :D this is the only necessary check. If both assets are 0, then asset_0_value becomes 0 and we will probably trigger a panic due to division by zero. So the check for both assets is needed.
The following 2 checks are not needed as the computation below handles these cases

lubkoll: and we should use checked operations probably

smart-contracts/osmosis/contracts/cl-vault/src/vault/deposit.rs
```diff
@@ -118,33 +93,42 @@ pub(crate) fn execute_any_deposit(
             (deposit_info.base_deposit, deposit_info.quote_deposit),
         ),
     )?;
-    let swap_calc_result = calculate_swap_amount(
-        deps,
+
+    // Get TWAP price
```
lubkoll, line 97: that comment is redundant. The function is already called get_twap_price.

smart-contracts/osmosis/contracts/cl-vault/src/vault/deposit.rs
```diff
@@ -61,10 +60,9 @@ pub(crate) fn execute_any_deposit(
         .position
         .ok_or(ContractError::MissingPosition {})?;
 
-    // get the amount of funds we can deposit from this ratio
-    // let (deposit_amount_in_ratio, swappable_amount): ((Uint128, Uint128), (Uint128, Uint128)) =
     let deposit_info = get_depositable_tokens(&deps.branch(), &info.funds, &pool_config)?;
 
+    // If we have no refunds let's proceed for the deposit
```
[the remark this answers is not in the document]
magiodev: idk, because of the early return. so to leave a clear feedback that this workflow ends here if that condition is met. i agree that the below code is explicative enough tho so we could avoid.

smart-contracts/osmosis/contracts/cl-vault/src/vault/deposit.rs
magiodev: yep indeed, sometimes i put TODO comments during the development and then i just remove the TODO leaving the useless comment there. I will remove that

smart-contracts/osmosis/contracts/cl-vault/src/helpers/getters.rs
magiodev: addressed, now we return Decimals here, which obbligates the invoker of this getter (working in Uint128) to then parse it. Its a bit worse there, but still better from a func perspective imo. Let me know if looks good to you

smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/autocompound.rs
```diff
@@ -268,71 +284,100 @@ fn test_autocompound_with_rewards_swap_non_vault_funds() {
...
+    // Get the contract balances for base and quote denoms after the swap of non vault funds
+    // Assert vault position tokens balances increased accordingly to the swapped funds from DENOM_REWARD to DENOM_BASE and DENOM_QUOTE
+    let balances_after_swap_base =
+        get_balance_amount(&app, contract_address.to_string(), DENOM_BASE.to_string());
+    // assert_approx_eq!(
+    //     expected_balance_base_after_deposit
+    //         .checked_add(expected_amount_rewards_to_base.into())
+    //         .unwrap(),
+    //     balances_after_swap_base,
+    //     &deposit_ratio_approx
```
magiodev, line 368: There are few assert_approx i had to comment out temporarily as the deposit_ratio_approx is not dynamic and doesnt allow to approx correctly. Numbers are correct tho. We need to find a way to assert that consistently.

smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/autocompound.rs
```diff
@@ -54,6 +56,8 @@ fn test_autocompound_with_rewards_swap_non_vault_funds() {
         )
         .unwrap();
 
+    // Get the initial vault token shares LP tokens supply right after the vault deployment
```
lubkoll, line 59: Could you remove this comment? In general we should try to comment only when absolutely necessary, as
* if they are necessary, then it might be better to refactor the code to be more readable
* often they comment the obvious (i.e. the naming in the following lines contains the same information as this comment)
* often they are misleading or wrong, the code is the real truth, so better let that document itself. I.e. here we the comment says we compare the burnt amount with the initial token supply, but the code shows that for some reason there is a factor 2 (that is the part that should be clarified as it looks sketchy)

smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/autocompound.rs
```diff
@@ -234,23 +241,32 @@ fn test_autocompound_with_rewards_swap_non_vault_funds() {
...
+
+    let contract_balance_base =
         get_balance_amount(&app, contract_address.to_string(), DENOM_BASE.to_string());
     assert_eq!(
         expected_balance_base_after_deposit.to_string(),
-        balances_base.to_string()
+        contract_balance_base.to_string()
     );
-    let balances_quote =
+    // for quote denom
```
lubkoll, line 259: rm?

smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/autocompound.rs
magiodev: you right, done!

smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/autocompound.rs
```diff
@@ -234,23 +241,32 @@ fn test_autocompound_with_rewards_swap_non_vault_funds() {
         shares_assets.balances[1].amount.u128()
     );
 
-    // Asssert contract balances for base and quote denoms
+    // Get the current contract balance again after deposits
+    // Asssert that contract balances for base and quote denoms are consistent with the amount of funds deposited and refunded by users
+    // for base denom
     let expected_balance_base_after_deposit = users_total_deposit_per_asset
-        .checked_add(balance_before_contract_base)
+        .checked_add(balance_before_contract_base) // this takes in account of the INITIAL_POSITION_BURN
```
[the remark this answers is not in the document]
magiodev: yes the comment was wrong, i also renamed all the variables using the convention you suggested

smart-contracts/osmosis/contracts/cl-vault/src/vault/swap.rs
```diff
@@ -123,75 +147,132 @@ pub fn execute_swap_non_vault_funds(
...
-            &pool_config.token0,
+    let pool_config = POOL_CONFIG.load(deps.storage)?;
+
+    // Determine the target token and amount based on swap direction
+    let (denom_out, amount_out_ratio) = match swap_direction {
+        SwapDirection::ZeroToOne | SwapDirection::AnyToOne => (
             &pool_config.token1,
-            token_in_amount
-                .checked_multiply_ratio(twap_price.numerator(), twap_price.denominator()),
+            (twap_price.numerator(), twap_price.denominator()), // TODO: Check if this is correct order for AnyToOne
```
magiodev, line 217: As in the previous two comments, we should check that the way we use twap_price here is consistent with how we inject arguments from above. The exact_deposit and any_deposit workflows are hardcoding the right assets on the requests, as we know we always work with asset0 and asset1 (base and quote). This is not the case for the SwapNonVaultFunds, which are accepting arbitrary routes and token ins from outside, and this could be hijackable.

magiodev:
Closing this in favor of https://github.com/quasar-finance/quasar/pull/782

PR #717, opened 2024-07-26
Remove code duplication in any_deposit and exact_deposit

lubkoll:


magiodev (commented):
Looking good! Made some suggestions in DM about how to better structure the execute_deposit function as private helper if we merge any_deposit.rs and exact_deposit.rs files maintaining the two distinct entrypoints.

magiodev (approved):
Approving as we are going to address structuring details in another separated PR.

PR #722, opened 2024-07-29
V3.0.0 - testnet version

arhamchordia:
## 1. Overview

This PR upgrades the Quasar chain from Cosmos SDK version 0.47 to version 0.50.x. This includes updating dependencies, modifying import paths, and adjusting the codebase to align with the changes and improvements introduced in the newer SDK version.

## 2. Implementation details
Ref for the cosmos-sdk upgrade - https://github.com/cosmos/cosmos-sdk/blob/release/v0.50.x/UPGRADING.md](https://github.com/cosmos/cosmos-sdk/blob/release/v0.50.x/UPGRADING.md

## 3. How to test/use

TODO. 

## 4. Checklist

- [ ] Does the Readme need to be updated? YES

## 5. Limitations (optional)

The upgrade may introduce breaking changes requiring dependent services or module modifications.
Extensive testing is needed to ensure compatibility with existing data and user workflows.


app/keepers/keepers.go
```diff
@@ -194,35 +198,40 @@ func (appKeepers *AppKeepers) InitNormalKeepers(
...
 	appKeepers.BankKeeper = bankKeeper
 
 	stakingKeeper := stakingkeeper.NewKeeper(
 		appCodec,
-		appKeepers.keys[stakingtypes.StoreKey],
+		runtime.NewKVStoreService(appKeepers.keys[stakingtypes.StoreKey]),
 		appKeepers.AccountKeeper,
 		appKeepers.BankKeeper,
 		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
+		address.NewBech32Codec(sdk.GetConfig().GetBech32AccountAddrPrefix()),
```
arhamchordia, line 227: Concern here as it gives cosmos as prefix instead of quasar

app/keepers/keepers.go
```diff
@@ -153,24 +156,25 @@ func (appKeepers *AppKeepers) InitSpecialKeepers(
...
 	appKeepers.CrisisKeeper = crisiskeeper.NewKeeper(
 		appCodec,
-		appKeepers.keys[crisistypes.StoreKey],
-		//app.GetSubspace(crisistypes.ModuleName),
+		runtime.NewKVStoreService(appKeepers.keys[crisistypes.StoreKey]),
 		invCheckPeriod,
 		appKeepers.BankKeeper,
 		authtypes.FeeCollectorName,
 		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
+		address.NewBech32Codec(sdk.GetConfig().GetBech32AccountAddrPrefix()),
```
arhamchordia, line 172: Gets codec as cosmos

app/keepers/keepers.go
ajansari95: setAddressPrefixes in config in  main.go before svrcmd.Execute

proto/quasar/tokenfactory/v1beta1/query.proto
```diff
@@ -0,0 +1,124 @@
...
+option go_package = "github.com/quasar-finance/quasar/x/tokenfactory/types";
+
+// Query defines the gRPC querier service.
+service Query {
+  // Params defines a gRPC query method that returns the tokenfactory module's
+  // parameters.
+  rpc Params(QueryParamsRequest) returns (QueryParamsResponse) {
+    option (google.api.http) = {
+      get: "/quasar/tokenfactory/v1beta1/params"
+      additional_bindings {
```
akure, line 19: Why is this additional binding required with the previous values proto paths? Is there any constrained do we have.

proto/quasar/tokenfactory/v1beta1/tx.proto
```diff
@@ -44,36 +47,46 @@ message MsgCreateDenomResponse {
...
-  [ (gogoproto.moretags) = "yaml:\"burn_from_address\"" ];
+      [ (gogoproto.moretags) = "yaml:\"burn_from_address\"" ];
 }
 
 message MsgBurnResponse {}
 
 // MsgChangeAdmin is the sdk.Msg type for allowing an admin account to reassign
 // adminship of a denom to a new account
 message MsgChangeAdmin {
+  option (amino.name) = "quasar/tokenfactory/change-admin";
```
akure, line 88: Why do we need the amino.name ?

proto/quasar/tokenfactory/v1beta1/query.proto
arhamchordia: @ajansari95

ajansari95: so we are backward compatible with old routes, if somebody has workflow it doesn't break. We can get rid of it later.Or if we dont care about backward compatibility we can remove.

ajansari95 (approved):
LGTM

akure (approved):

akure:
Sounds good to me.

PR #731, opened 2024-07-30
Remove commented out contracts

lubkoll:


arhamchordia (commented):
Again same question here that we should remove these or not. I don't think we should remove airdrop contract.

LaurensKubat:
Imo all these can be removed. If we need them again we can just readd them later. for the airdrop contract, we'd probably reuse our merkle incentives contract and infra right now to do a merkle airdrop

LaurensKubat (dismissed):
lgtm

ajansari95 (approved):
lgtm

PR #732, opened 2024-07-30
Remove commented out packages

lubkoll:


arhamchordia (approved):
I don't have much context on these packages removal. But if this has been discussed then it looks good to me. 
Can be after an approval from @LaurensKubat

PR #740, opened 2024-08-05
Fix/proto

ajansari95:
WIP


app/upgrades/v3/constants.go
```diff
@@ -0,0 +1,27 @@
...
+	crisistypes "github.com/cosmos/cosmos-sdk/x/crisis/types"
+	icqtypes "github.com/cosmos/ibc-apps/modules/async-icq/v8/types"
+	ibcwasmtypes "github.com/cosmos/ibc-go/modules/light-clients/08-wasm/types"
+	"github.com/quasar-finance/quasar/app/upgrades"
+)
+
+// UpgradeName defines the on-chain upgrade name for the Quasar chain v2.0.0 upgrade.
+const UpgradeName = "v3"
+
+var Upgrade = upgrades.Upgrade{
```
arhamchordia, line 15: Just add a todo here. In case we forget about circuit types.

arhamchordia (commented):
In general it looks good. But needs some cleanup in imports and address build issue.

arhamchordia:
In my local reproducible fails with wasm binding issue.  Which can be addressed in another PR open against this branch itself.

arhamchordia (approved):
We deal with issues in a different PR.

PR #746, opened 2024-08-07
Fixing export, some query commands, autocli imports and so on.

arhamchordia:


app/export.go
```diff
@@ -115,40 +123,65 @@ func (app *QuasarApp) prepForZeroHeightGenesis(ctx sdk.Context, jailAllowedAddrs
 	height := ctx.BlockHeight()
 	ctx = ctx.WithBlockHeight(0)
 
+	// reinitialize all validators (v0.46 version)
```
ajansari95, line 126: do we need this commented part?

arhamchordia: Added it just in case to keep it as a reminder to check back on it later on.

ajansari95 (approved):

PR #753, opened 2024-08-08
Separate position creation from swap

lubkoll:


smart-contracts/osmosis/contracts/cl-vault/src/vault/range.rs
```diff
@@ -217,58 +259,18 @@ pub fn do_swap_deposit_merge(
...
-        POSITION.save(
-            deps.storage,
-            &Position {
-                position_id: position_id.expect("position id should be set if no swap is needed"),
-                join_time: position.join_time,
-                claim_after: position.claim_after,
-            },
-        )?;
-
-        SWAP_DEPOSIT_MERGE_STATE.remove(deps.storage);
```
LaurensKubat, line 268: We should still make sure to clean up our state item after we're done using it

LaurensKubat (changes requested):
generally lgtm, but we should remove  SWAP_DEPOSIT_MERGE_STATE when we're done using it, last remove was deleted as part of the refactor. Also I think this removes our ability to just open a position. But that's probably for a different PR

smart-contracts/osmosis/contracts/cl-vault/src/vault/range.rs
```diff
@@ -195,13 +203,47 @@ pub fn handle_initial_create_position_reply(
 
     let unused_pair_balances = get_unused_pair_balances(&deps, &env, &pool_config)?;
 
+    if unused_pair_balances
```
LaurensKubat, line 206: This if condition means that we still have to swap some amount of tokens, combined with the middleware, we always swap 100% of the swappable tokens and thus cannot do easy inventory management. I don't think we should fix that for this PR

smart-contracts/osmosis/contracts/cl-vault/src/vault/range.rs
lubkoll: The cleanup happens in `handle_iteration_create_position_reply`. Here it was just deleted as it was written before without needing it.

lubkoll:
> generally lgtm, but we should remove SWAP_DEPOSIT_MERGE_STATE when we're done using it, last remove was deleted as part of the refactor. Also I think this removes our ability to just open a position. But that's probably for a different PR

Shouldn't have changed anything in the logic, just a refactor to separate concerns

smart-contracts/osmosis/contracts/cl-vault/src/vault/range.rs
LaurensKubat: Sounds good, misread then

smart-contracts/osmosis/contracts/cl-vault/src/vault/range.rs
LaurensKubat: Alright, think I misunderstood then

LaurensKubat (approved):
lgtm

PR #755, opened 2024-08-08
WIP: Cleanup rebalance implementation

lubkoll:
This one needs a thorough review. I removed a lot of stuff, so there is a higher chance that I missed something.


LaurensKubat (commented):
This looks to keep the same total functionality as before but cleaner

lubkoll:
This is currently blocked by the cl-vault state. After enough cleanup, there will be a time where we can write tests. Will rewrite then.

PR #769, opened 2024-08-09
WIP: Introduce PoolPair and remove assert.rs

lubkoll:


smart-contracts/osmosis/packages/quasar-types/src/pool_pair.rs
```diff
@@ -0,0 +1,10 @@
+pub struct PoolPair<T> {
```
LaurensKubat, line 1: Changing this struct to take two generic parameters would allow users of this struct to use type aliases for their denoms, for example 
```
pub struct PoolPair<B, Q> {
    pub base: B,
    pub quote: Q,
}

type Osmo = String;
type Quasar = String;

let qsr_pool = PoolPair<Osmo, Quasar>
```
 Similar to https://github.com/EntropicLabs/cw-safety/blob/main/crates/monetary/src/rate.rs, and would also allow PoolPair to be used in more different implementations
 
 Actual usage in the contract of the wider generic implementation can come later, for now we can just use `PoolPair<Coin, Coin>`

LaurensKubat (changes requested):
One comment on the usage of generics in PoolPair, other than that lgtm

smart-contracts/osmosis/packages/quasar-types/src/pool_pair.rs
lubkoll: I am not convinced by that as:
1. We don't need to have two generic parameters now, so we would spend work on a potential generalization that has a high chance of never being used.
2. Using type aliases to rename types is often bad. It makes a lot of sense for types that are the result of specializing a complex generic (i.e. such as often done with ContractResult). Renaming types in contrast often just makes it harder to understand what types are actually used and undermines type safety as all these different types are actually not different.

LaurensKubat: In that case, my next question would why we would make PoolPair have a generic at all. I'd say that we should either go for very flexible or not flexible and just use Coin in the Pool. Having just one generic parameter leads to us not really using it in my mind, because with on generic parameter you can't use it to add extra type safety or add any extra functionality through the type system. This implementation would allow PoolPairs to be used for Coin<>Coin or CW20<>CW20 pairs, but not any mixed pools or something similar

LaurensKubat: As for 2. We can also alias to add data, such as `type Quasar = Coin`, where the String field is always set to the correct denom of QSR, For crosschain tokens that becomes a bit more complex. 
My main idea here is to enable the PoolPair to be used with more complex generics later, without forcing a rewrite of Poolpair

LaurensKubat: Also my above Alias example is wrong, this would be the correct way to do it most likely: https://github.com/EntropicLabs/cw-safety/blob/f7c0f9ccefa5a39706376306eeea7b202470b96c/crates/monetary/src/payment.rs#L59

lubkoll: Moved the whole PR into WIP. Imo we should have a short discussion about this to decide on the which way to move forward, but only when all the other cl-vault stuff is done. Until then I prefer to park this.

LaurensKubat: Sounds good, lmk when

lubkoll:
Closing this as improved versions of the changes are on main or will be part of new PRs

PR #785, opened 2024-08-19
Move handle_merge_reply into autocompound.rs

lubkoll:
.. as that is the place where the corresponding SubMsg is created


ajansari95 (approved):
lgtm

PR #788, opened 2024-08-20
Simplify ModifyRange implementation

lubkoll:


smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/range.rs
```diff
@@ -51,7 +51,7 @@ fn move_range_works() {
             &ExecuteMsg::VaultExtension(ExtensionExecuteMsg::ModifyRange(ModifyRangeMsg {
                 lower_price: Decimal::from_str("0.65").unwrap(),
                 upper_price: Decimal::from_str("1.3").unwrap(),
-                max_slippage: Decimal::bps(MAX_SLIPPAGE_HIGH),
+                max_slippage: Decimal::percent(89),
```
LaurensKubat, line 54: nitpick: A comment why 89% is appropriate in this test is good imo, since it's not directly a logical value

LaurensKubat (approved):
In general this looks good to me, except that all functions use 18 digit decimals, thus small amounts of 18 decimal tokens at low spot prices are not supported, not needed for this PR but something we should track and fix after

smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/range.rs
lubkoll: :D something about the test setup. With 90% it failed with a slippage error that missed by one micro-unit

PR #792, opened 2024-08-20
Clone element instead of vector

lubkoll:


lubkoll:
whoever approves, feel free to also merge

magiodev (approved):
Good catch, looks good!

PR #793, opened 2024-08-20
Replace PoolAssets with PoolPair

lubkoll:


lubkoll:
might be worth merging after the swap non-vault funds. There shouldn't be a merge conflict, but it has some overlap that could get git confused

LaurensKubat (approved):
lgtm

PR #782, opened 2024-08-19
CL: Swap Non Vault Funds

magiodev:
## 4. Checklist

- [ ] Does the Readme need to be updated?

PR #802, opened 2024-08-21
CL: Swap Non Vault Funds (#782)

magiodev:


smart-contracts/osmosis/contracts/cl-vault/src/vault/swap.rs
```diff
@@ -85,6 +77,42 @@ pub fn execute_swap_non_vault_funds(
...
+    forced_swap_route: Option<Vec<SwapAmountInRoute>>,
+    twap_window_seconds: Option<u64>,
+) -> Result<CosmosMsg, ContractError> {
+    let vault_config = VAULT_CONFIG.load(deps.storage)?;
+    let dex_router = DEX_ROUTER.may_load(deps.storage)?;
+
+    let twap_price = get_twap_price(
+        &deps.querier,
+        env.block.time,
+        twap_window_seconds.unwrap_or_default(),
```
lubkoll, line 95: Is that correct, i.e. in case we don't provide that from the outside, then we have a twap_window of 0 now. Shouldn't that also be some default-value (like the 24 hardcoded in range.rs)?

smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/autocompound.rs
```diff
@@ -82,21 +82,22 @@ fn test_autocompound_with_rewards_swap_non_vault_funds() {
...
+    // BEFORE USERS DEPOSITS
 
     // Keep track of the total refunded amount on token0 from user deposits
     let mut refund0_amount_total = Uint128::zero();
     let mut refund1_amount_total = Uint128::zero();
-
     // Keep track of the total minted shares from deposits
     let mut total_minted_shares_from_deposits = Uint128::zero();
+
+    // This will be useful after all deposits to assert the contract balance
```
lubkoll, line 93: plz rm comment, that does not help reading the test code

smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/autocompound.rs
```diff
@@ -266,98 +268,113 @@ fn test_autocompound_with_rewards_swap_non_vault_funds() {
...
+        initial_rewards_balance.to_string(),
     );
 
     // SWAP NON VAULT ASSETS BEFORE AUTOCOMPOUND ASSETS
 
-    // Swap non vault funds to vault funds
-    // 50000000000ustride to 49500000000uatom as spot price 1.0 less swap_fees
-    // 50000000000ustride to 49500000000uosmo as spot price 1.0 less swap_fees
-    let expected_amount_rewards = 49500000000u128;
+    // Calculate the total rewards swap amount using balanced_fee configs (0.01 is coming from osmosis as flat fee)
```
lubkoll, line 280: What is balanced_fee configs? And why is it important that 0.01 is a "flat" fee?
to me the code after is clearer than the comment which leads to questions that are not relevant for this test case. I'd prefer to remove that

smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/autocompound.rs
```diff
@@ -366,24 +383,19 @@ fn test_autocompound_with_rewards_swap_non_vault_funds() {
...
-        expected_refund_base,
-        balances_after_autocompound_base,
-        &deposit_ratio_approx
-    );
-    // Quote about rewards swapped to token quote
-    let balances_after_autocompound_quote =
+    let after_autocompound_quote_balance =
         get_balance_amount(&app, contract_address.to_string(), DENOM_QUOTE.to_string());
-    assert_eq!(expected_refund_quote, balances_after_autocompound_quote);
+    // Assert that either the base balance is 0, or the quote balance is 0, or both are 0.
```
lubkoll, line 390: that is easy to understand from the following line. though the comment is confusing as it explicitly mentions the case that both are 0, which is already handled and thus not checked for in the implementation

smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/setup.rs
```diff
@@ -585,7 +587,7 @@ pub fn calculate_deposit_ratio(
     };
 
     // TODO: Compute this based on tokens_provided size
-    let ratio_approx: String = "0.00005".to_string();
+    let ratio_approx: String = "0.02".to_string();
```
lubkoll, line 590: I recall we spoke about this once, but nothing more. What's the reason for this big change?

lubkoll (changes requested):

smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/autocompound.rs
magiodev: Is a flat fee on test_tube https://github.com/osmosis-labs/test-tube/blob/main/packages/osmosis-test-tube/src/module/gamm.rs#L59 as the Gamm module from it is hardcoding 0.01% swap fee and we cannot change that.

smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/setup.rs
magiodev: I've been able to revert this to the original value after splitting 50% 50% and tweaking a bit the deposit values. I also deprecated it from the setup.rs and inlined it in the autocompounds.rs which is the only place that needs it.

smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/autocompound.rs
```diff
@@ -277,16 +270,16 @@ fn test_autocompound_with_rewards_swap_non_vault_funds() {
 
     // SWAP NON VAULT ASSETS BEFORE AUTOCOMPOUND ASSETS
 
-    // Calculate the total rewards swap amount using balanced_fee configs (0.01 is coming from osmosis as flat fee)
+    // Calculate the total rewards swap amount using 0.01 as swap_fee as osmosis_test_tube hardcodes it this way: https://github.com/osmosis-labs/test-tube/blob/main/packages/osmosis-test-tube/src/module/gamm.rs#L59
```
lubkoll, line 273: fine if you keep it. I still think it is redundant. Of course we have to take into account the fees as osmosis does

smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/autocompound.rs
```diff
@@ -322,27 +315,24 @@ fn test_autocompound_with_rewards_swap_non_vault_funds() {
...
     assert_eq!(0u128, after_swap_rewards_balance);
 
     let after_swap_base_balance =
         get_balance_amount(&app, contract_address.to_string(), DENOM_BASE.to_string());
     // Assert vault position tokens balances increased accordingly to the swapped funds from DENOM_REWARD to DENOM_BASE and DENOM_QUOTE
-    assert_approx_eq!(
+    assert_eq!(
         after_deposit_base_balance
-            .checked_add(rewards_swap_amount.into())
+            .checked_add(swap_token_in_amount.into())
```
lubkoll, line 325: that looks now totally wrong. After a swap I expect `before_swap_balance - swap_token_in_amount = after_swap_balance` and `before_swap_balance + swap_out_amount = after_swap_balance`

smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/autocompound.rs
```diff
@@ -134,9 +130,8 @@ fn test_autocompound_with_rewards_swap_non_vault_funds() {
                     .unwrap()
                     .u128(),
                 expected_refund0,
-                &deposit_ratio_approx
+                APPROX_EQ_FACTOR
```
lubkoll, line 133: fine for now, but it would have been better to do that in a separate PR. Here it is not necessary and added another 8 files to the PR

lubkoll (changes requested):

smart-contracts/osmosis/contracts/cl-vault/tests/test-tube/autocompound.rs
magiodev: Yes you right, the math was right as here, working with the assump[tion of spot price = 1.0 we could do that. But the nomenclature was clearly wrong. Fixed!

lubkoll (dismissed):

smart-contracts/osmosis/contracts/cl-vault/src/contract.rs
```diff
@@ -214,8 +216,27 @@ pub fn reply(deps: DepsMut, env: Env, msg: Reply) -> Result<Response, ContractEr
 }
 
 #[cfg_attr(not(feature = "library"), entry_point)]
-pub fn migrate(deps: DepsMut, _env: Env, _msg: MigrateMsg) -> Result<Response, ContractError> {
-    MIGRATION_STATUS.save(deps.storage, &MigrationStatus::Open)?;
+pub fn migrate(deps: DepsMut, _env: Env, msg: MigrateMsg) -> Result<Response, ContractError> {
```
LaurensKubat, line 219: Add a unit test for this behaviour in the migration please, this also helps in clarifying for the next rollout what you expect to happen here

LaurensKubat (dismissed):
lgtm

smart-contracts/osmosis/contracts/cl-vault/src/contract.rs
magiodev: Done!

lubkoll (approved):

PR #806, opened 2024-08-22
[CI] Add cachepot -- part 1: generate cache

lubkoll:


magiodev (approved):
Changes looks good, but I didn't test the CI execution part. I guess we can merge and see if everything executes correctly.

PR #813, opened 2024-08-23
Cleanup github runner at start of all_checks_osmosis

lubkoll:


.github/workflows/all_checks_osmosis.yml
```diff
@@ -25,6 +25,8 @@ jobs:
         uses: actions/checkout@v3
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
+      - name: Cleanup runner
+        run: sudo rm -rf /usr/local/share/boost && sudo rm -rf /usr/share/dotnet && sudo rm -rf /usr/local/lib/android && sudo rm -rf /opt/ghc && sudo rm -rf /opt/hostedtoolcache/CodeQL && sudo docker image prune --all --force
```
magiodev, line 29: Why .NET and Android references? Is that something we expect to find during the cleanup?

magiodev, line 29: Also, would sudo `docker system prune` be better in this context? So we also remove possible unused networks, volumes, etc.

lubkoll: yes, github installs stuff that is used often into the runners by default. This cleans up more than 25GB of disk space

lubkoll: tried that too, didn't make a difference

magiodev (approved):

PR #817, opened 2024-08-26
Reduce redundancy in denom-checks

lubkoll:


magiodev (approved):
Nice cleanup, approved

PR #820, opened 2024-08-26
Remove PR template

lubkoll:


arhamchordia (commented):
I have been using this template for bigger PRs of chain. I don't see point in removing it. 
We can totally refine this but removing imo is not necessary.

ajansari95:
> I have been using this template for bigger PRs of chain. I don't see point in removing it. We can totally refine this but removing imo is not necessary.

second this on chain side there are cases where we change a lot of stuff this comes in handy to get context for reviewer.

PR #822, opened 2024-08-26
Use dex-router in test-tube tests

lubkoll:
This is part of the dex-router removal. To reduce the scope of that, I changed some of the test-tube tests to actually use the dex-router.


ajansari95 (commented):
lgtm , but any motivation behind changing the hardcoded test cases  values (which would have been avoided earlier IMO ),  is it due to any logical changes that we made earlier?

lubkoll:
> lgtm , but any motivation behind changing the hardcoded test cases values (which would have been avoided earlier IMO ), is it due to any logical changes that we made earlier?

Yes, absolutely. The dex-router was never used. I didn't want to change to much of the setup just to get to the same values as before.

ajansari95 (approved):

PR #823, opened 2024-08-26
Remove authz messages

lubkoll:


magiodev (approved):
changes look good to me. please ensure that we internally confirm whether this implementation was done for a specific reason or to support any particular consumer. this has been kept unused so far afaik tho

PR #825, opened 2024-08-26
Remove unnecessary dependencies from get_unused_(pair_)balances

lubkoll:


smart-contracts/osmosis/contracts/cl-vault/src/helpers/getters.rs
```diff
@@ -202,20 +202,30 @@ pub fn get_single_sided_deposit_1_to_0_swap_amount(
...
+pub fn get_unused_pair(
+    deps: &Deps,
+    addr: &Addr,
+    pool_config: &PoolConfig,
+) -> Result<PoolPair<Coin, Coin>, ContractError> {
+    let unused_balances = get_unused_balances(&deps.querier, addr)?;
+    Ok(get_vault_funds_or_zero(&unused_balances, pool_config))
+}
+
 pub fn get_unused_pair_balances(
```
magiodev, line 223: Changes mostly look good. I noticed a minor redundancy issue with the get_unused_pair and get_unused_pair_balances functions. both functions seems to perform the same task of returning the unused funds related to the pair that the vault is managing. assuming addr is always the contract address one or the other becomes redundant. am i missing something?

lubkoll: No, you are right. The motivation for that is not clear from this PR, probably should have added a comment for that.
This is sth. I found when doing the dex-router PR and that I could extract into a separate PR. There I will use get_unused_pair. I'd expect that we will transition from using the function that returns a vector to the one that uses a PoolPair. As `get_unused_balances` is used a lot, this makes it easier to iterate on that with smaller PRs.

lubkoll: ticket for cleanup: https://app.clickup.com/t/8695hgebd

magiodev (approved):

PR #843, opened 2024-09-03
Remove SCM.md

lubkoll:


lubkoll:
We don't do what's described in this doc and I can't see how it adds value

magiodev (approved):

